PR #25, opened 2026-07-23
Feature/Fix: support GLM for all engines mix and vllm moriio pd & adding all to test

dorado269:
# Description

Fix the vLLM v0.25.1 MoRIIO PD-disaggregation corruption on block-scaled fp8 MLA models (DeepSeek-V4 / GLM-5.1), and add automated GLM-5.1 e2e coverage across **all three engines** (vLLM, SGLang, ATOM) so it can't regress.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Infra/Build change
- [x] Code refactoring

## Changes

1. **Remove three no-op MoRIIO patches** — `patch_moriio_dsa_write.py`, `patch_moriio_hetero.py`, `patch_vllm_moriio_blocksize.py` targeted the pre-v0.25.1 connector. v0.25.1 refactored it into `moriio_layout.py`, which natively covers all three concerns (full-layer `wait_for_save`, per-layer transfer geometry, kernel/logical block-size ratio), so their anchors no longer match and each silently skips. Dropped to remove dead weight + misleading "skipped" build-log lines.

2. **Fix the MLA page-length transfer geometry** — the real bug. `get_layer_transfer_geometry` derived the MLA per-block transfer size/stride from tensor shape (`block_size*latent*element_size`) instead of the authoritative `spec.page_size_bytes`. Two symptoms, one defect:
   - **DeepSeek-V4** `fp8_ds_mla` is 576-byte-aligned/padded → `page_size_bytes` > shape bytes; the dropped tail carries the UE8M0 per-block scale → decode dequantizes with a stale scale → right-structure/wrong-fact output ("France"→"a good idea").
   - **GLM-5.1** lays the cache out per kernel block of size 1 while the scheduler pages at block_size 16 → shape-derived length is 16× too small → only 1/16 of each block's KV moved → total garbage ("is is is").

   `patch_moriio_pagelen.py` uses `spec.page_size_bytes` for `block_len` and `page_size_bytes // element_size` for `block_stride`, mirroring Mooncake (correct on the same nodes). Idempotent, self-locating, no-op for contiguous fp16/bf16 K/V.

3. **Add GLM-5.1-FP8 e2e cases across all engines/scenarios** — the regression guards. For vLLM PD-disagg, the adapter now selects the KV connector per-case: a case opts into MoRIIO via `extra_env[INFERA_E2E_KV_CONNECTOR]="MoRIIOConnector"`, otherwise it **defaults to Mooncake so every existing case is byte-for-byte unchanged**. Only GLM declares the sentinel; no other model gets a MoRIIO variant. vLLM/SGLang/ATOM also get single-node PD-mixed GLM cases via their existing adapters (no adapter change). All e2e cases stay `@pytest.mark.slow` (skipped by default) → CI is unchanged.

## Engine / case coverage (GLM-5.1-FP8)

GLM-5.1-FP8 was brought up and verified correct at temp=0 (France→Paris, China→Beijing, 2+2→4) on every engine/scenario, and each is wired into the suite:

| Engine | Scenario | KV transfer | TP | e2e case (test id) | Live temp=0 verify |
|--------|----------|-------------|----|--------------------|--------------------|
| vLLM   | **PD-disaggregated** (cross-node) | **MoRIIO** (this PR's fix) | 4 | `pd_disag/vllm` → `GLM-5.1-FP8-tp4` | ✅ PASS (2-node) |
| vLLM   | PD-mixed (single-node) | n/a (in-process) | 4 | `pd_mixed/vllm` → `GLM-5.1-FP8-tp4` | ✅ PASS |
| SGLang | PD-mixed (single-node) | n/a (in-process) | 4 | `pd_mixed/sglang` → `GLM-5.1-FP8-tp4` | ✅ PASS |
| ATOM   | PD-mixed (single-node) | n/a (in-process) | 4 | `pd_mixed/atom` → `GLM-5.1-FP8-tp4` | ✅ PASS |

Existing (unchanged) transfer-engine coverage for context: vLLM `pd_disag` also runs **Mooncake** (default) on `Qwen3-0.6B-tp2`; the MoRIIO path is new and GLM-only.

Notes per engine:
- **vLLM/MoRIIO** (pd_disag) is the only case exercising this PR's page-len fix (cross-node RDMA KV transfer). Garbage before the fix, correct after. The vLLM pd_mixed case uses no kv-transfer connector, so the fix is a no-op there — it's a straight brings-up-and-is-correct guard.
- **SGLang**: `GlmMoeDsaForCausalLM` auto-selects the DSA attention path (`attention_backend=dsa`, tilelang); the case is minimal (`--reasoning-parser glm45`, `SGLANG_USE_AITER=1`) — forcing DSv4 flags would fight the auto-config.
- **ATOM**: loads `GlmMoeDsaForCausalLM` natively (MLA chunked-prefill workspaces); minimal `--kv_cache_dtype fp8 --trust-remote-code`. **No MTP** — GLM ships no MTP/nextn draft weights and gfx950 plain decode is correct (the gfx942 broken-plain-decode bug does not reproduce).

## Verification

- **Image build** (cluster, MI355X node): `docker build -f deploy/docker/Dockerfile.vllm` succeeds; build log shows `moriio-pagelen: patched` (not skipped) and the three removed patches never appear; in-image `moriio_layout.py` contains the fix marker.
- **Fix correctness** (TP4 2-node MoRIIO PD, temp=0): DeepSeek-V4-Pro and GLM-5.1-FP8 both France→Paris / China→Beijing, coherent and deterministic. Mooncake was correct on the same nodes throughout (differential reference).
- **Test wiring** (`pytest --collect-only`, offline):
  - `pd_disag/vllm` → `Qwen3-0.6B-tp2` (existing, unchanged) + `GLM-5.1-FP8-tp4` (new MoRIIO). Adapter unit assertions confirm the Mooncake path argv/env is byte-for-byte unchanged and the MoRIIO path emits the correct extra_config with the sentinel stripped from worker env.
  - `pd_mixed/vllm` → existing 3 ids unchanged + `GLM-5.1-FP8-tp4`.
  - `pd_mixed/sglang` → existing 4 ids unchanged + `GLM-5.1-FP8-tp4`.
  - `pd_mixed/atom` → existing 4 ids unchanged + `GLM-5.1-FP8-tp4`.

## Test plan

- [x] Image builds with the fix baked in via the Dockerfile patch loop
- [x] `pytest --collect-only` lists the new cases; existing ids / default Mooncake path unchanged
- [x] GLM-5.1-FP8 temp=0 correctness verified live on all four engine/scenario combos (vLLM MoRIIO 2-node, vLLM single-node mix, SGLang single-node, ATOM single-node)
- [ ] (optional, needs 2 nodes) `INFERA_E2E_NODES=<p>,<d> tests/run_tests.sh e2e vllm disag` → GLM MoRIIO correctness probe PASS in-suite

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jiejingzhangamd:
we add DOC in release check, so please add sign off in git commit.

jiejingzhangamd (approved):

PR #34, opened 2026-07-28
sglang DSA: enable PD + DP-attention + EAGLE MTP on GLM-5.2 / gfx950 (4 patches)

dorado269:
Four patches against the sglang source tree bundled in the ROCm engine images
(`lmsysorg/sglang:v0.5.15.post1-rocm720-mi35x` and derivatives, where sglang is an
editable checkout at `/sgl-workspace/sglang`, commit
`0b3bb0cbe31873994c9f989fddfe2f87ca839fdd`).

Together they make **PD disaggregation + DP-attention + EAGLE MTP** work on GLM-5.2-MXFP4 /
gfx950, with the draft CUDA graph enabled. Each one is independently necessary — without
any single one of them the configuration crashes or deadlocks.

> **These diffs are NOT applied at image build time.** `Dockerfile.sglang` no longer
> carries a patch loop (the previous `deploy/docker/patches/sglang/*.py` mechanism was
> removed along with it), so a stock build of the image contains none of them. Every result
> below was obtained by applying them into a running container. Baking them in would need a
> build step; that is deliberately not done here, since they target a specific pinned sglang
> commit and would silently no-op or conflict on a different base.

## The patches

| # | file | fixes | symptom without it |
|---|---|---|---|
| 1 | `dsa_indexer_hip_dp_padded_rows.diff` | HIP indexer sizes `logits` from DP-**padded** rows while `lengths` is sized to real rows | `RuntimeError: Expected lengths.size(0) == B to be true, but got false` on the first batch |
| 2 | `dsa_backend_dp_sync_and_page_table_rows.diff` | (a) rank-divergent GPU→CPU syncs; (b) page-table rows (per request) vs top-k rows (per token) under MTP | (a) PD+MTP deadlock; (b) `assert page_table.shape[0] == topk_indices.shape[0]` under concurrency |
| 3 | `deepseek_nextn_glm52_mtp_bf16.diff` | GLM-5.2's nextn layer is bf16 while the rest of the model is quantized | dies at weight load: `size of tensor a (3072) must match ... b (6144)` |
| 4 | `eagle_worker_v2_uniform_draft_graph.diff` | the draft graph/eager decision is made per rank from rank-dependent inputs | first routed request hard-deadlocks on the PD decode leg |

### 1 — HIP indexer, padded vs real rows

`dsa_indexer.py::_get_topk_paged`: the CUDA path slices its paged-MQA inputs to the real
(unpadded) row count `q_offset` — which is what `lengths` is sized to — then restores the
padding. The aiter/HIP path did neither: it sized `logits` from the DP-padded
`q_fp8.shape[0]`, and the padding-restore was gated `not _is_hip`. Under DP-attention the
hidden states are padded, so top-k saw `score.shape[0] != lengths.shape[0]` and asserted.
The patch makes HIP follow the CUDA contract. Two guards keep the non-DP path
bit-identical.

The slice guard was originally `if 0 < q_offset < q_fp8.shape[0]`; the `0 <` lower bound
made the slice a no-op on DP-**idle** ranks (`q_offset == 0`, the rank holds no requests),
reproducing the same assert there. That only surfaces once some ranks are idle while others
are busy — i.e. under concurrency, which the original single-node validation never
exercised. The bound is removed: `q_fp8[:0]` is a legal empty slice and is what an idle rank
should pass, and the CUDA path has no such bound either.

Not the same defect as sglang PR #30378 / #30427, which clamp padded row *values*
(`seqlens_expand_kernel`, already present in this image). This fixes the HIP-side row
*count*.

### 2 — rank-divergent host syncs, and page-table rows

**(a)** `max_seqlen_k = int(forward_batch.seq_lens.max().item())` is a host sync sitting on a
branch only *some* DP ranks take; idle peers take the cheap arm, the DP collectives
desynchronize, the group deadlocks. Replaced with the sync-free
`self.req_to_token.shape[1]` — the same idiom `_graph_page_table_width()` and the
graph-capture paths already use. Over-allocating columns is safe: the page table is only
indexed through top-k, which masks per row by `cache_seqlens`. Two further unconditional
`.cpu()` syncs are removed.

**(b)** `metadata.page_table_1` is `req_to_token[req_pool_indices]`, one row per **request**,
while `topk_indices` has just been padded to `q.shape[0]`, one row per **token**. Equal for
plain decode, not under MTP where the draft model runs several tokens per request. A helper
expands the page table to token rows via `repeat_interleave`. Rows added by
`_pad_topk_indices` are all `-1` and are masked out by the triton kernel
(`valid_topk_mask = mask & (loaded_topk_indices >= 0)`), so only the row *count* has to
match. Applied at **both** decode call sites — the traceback named only the first, but the
second pairs the same unpadded page table with a padded `topk_indices`.

### 4 — uniform draft-graph decision (the PD deadlock)

`eagle_worker_v2.py::draft()` decides per rank whether to replay the draft CUDA graph or run
the multi-step draft eagerly. The two paths do not issue the same host-side collective
sequence, and two of the guard's four terms are rank-dependent **by construction**:

* `not forward_batch.forward_mode.is_idle()` — occupancy differs per rank every step;
* `draft_input.dsa_topk_indices is None` — on the **PD decode leg** this field is built from
  RDMA-shipped per-request payloads (`eagle_disaggregation.py:54-59`), so it is a function of
  which requests a rank happens to hold.

So one rank replays a graph while another runs eager inside the same collective step, and
the group desynchronizes permanently.

This was measured, not argued. Instrumenting all four terms per rank per iteration: the
*decision* diverged on **exactly** the iteration that froze — one busy rank eager, seven idle
ranks graph — and that rank is exactly the one py-spy found blocked in the eager
`init_forward_metadata`. Reproduced twice with different victim ranks (dp2, dp3), which is
the race behaving as a race. Divergence in the individual terms is routine and usually
cancels; it is fatal only when it reaches the final decision, which is why warmup passes.

This also explains why single-node **mix** never hangs: with no disaggregation the top-k seed
is always produced locally by identical code on every rank, so the fourth term never flips
asymmetrically.

The fix votes the local need-for-eager across the DP group (one 1-element **gloo** all-reduce
on `cpu_group` — no GPU sync, cannot serialize the compute stream) and has every rank act on
the group's answer. An idle rank contributes `False`, so it can never drag the group to
eager.

Two details that would otherwise make the patch silently inert:

* it must use `get_tp_group()`, **not** `get_attention_tp_group()` — under DP-attention with
  `dp_size == tp_size == 8` the attention TP group is `attn_tp_size = tp/dp = 1` rank wide, so
  an all-reduce over it returns the local value and the fix compiles, runs, and changes
  nothing. The group that must agree is the full TP group, the same one the scheduler's
  MLP-sync all-gather spans (`dp_attn.py:91`);
* a collective is safe at this line because instrumentation showed all 8 ranks entering
  `draft()` an equal number of times every iteration, idle ones included — the vote sits on a
  branch every rank takes, which is precisely the property the buggy code lacked.

An earlier attempt that removed only the `not is_idle()` term does **not** work, and the
measurements say why: replaying the recorded iterations, it fixes the busy-rank one and
breaks the idle-rank one — matching the observed symptom of the failure moving *earlier*,
into warmup.

## Verification

**Config:** 2× 8×MI355X (gfx950), ROCm/HIP 7.2.26015, torch 2.9.1+rocm7.2.0, sglang
0.5.15.post1, GLM-5.2-MXFP4, PD over mooncake RDMA (mlx5_0 + dma-buf + GID 3),
`--tp-size 8 --dp-size 8 --enable-dp-attention --ep-size 8`, EAGLE MTP (steps=3, topk=1,
draft-tokens=4), **draft CUDA graph enabled**.

| Check | Result |
|---|---|
| PD warmup (8 concurrent, one per DP rank) | pass |
| conc 1/2/4/8/16/64/128/256 × 512 tokens | **1384/1384**, 0 failures, 0 KVTransferError |
| Single-node mix regression (shared code path) | **132/132** |
| Durability, 8 × conc=128 × 512 back to back | **1024/1024**, accept length flat at ~3.0 |
| **Draft graph actually used** | **98.4%** of iterations |
| Same-node control, only patch 4 reverted | **0/4 — deadlock, 120 s timeout on request 1** |

Cumulative **2540/2540** with the patches, **0/4** without patch 4.

The mechanism was verified, not just the outcome. Over 2992 iterations with all 8 ranks
logging: the *local* decision still diverges 38 times (the defect is latent and real, not
masked by timing) while the *acted-on* decision diverges **0** times, and the vote changed
some rank's mind 190 times — each one an averted deadlock. The 98.4% graph usage is what
distinguishes this from simply disabling the draft graph, which was the previous workaround.

Patch 1 was separately verified single-node (`--dp-size 8 --enable-dp-attention --ep-size 8`
+ EAGLE): 4/4 correctness, accept length median 3.86/4 (n=251), conc=64 ISL/OSL 1k/1k
256/256 requests 0 failed, median TPOT 17.2 ms; DP-attention-only and MTP-only regressions
unchanged.

## Applying

All four apply with exact context, no fuzz:

```bash
cd /sgl-workspace/sglang
for d in dsa_indexer_hip_dp_padded_rows.diff \
         dsa_backend_dp_sync_and_page_table_rows.diff \
         deepseek_nextn_glm52_mtp_bf16.diff \
         eagle_worker_v2_uniform_draft_graph.diff; do
  patch -p1 --fuzz=0 < "$d"
done
```

Verified by applying for real into a scratch tree and byte-compiling the result — note that
`patch --dry-run` and `git apply --check` **fuzz by default**, and a hand-written diff in
this series once silently dropped a hunk while still "passing".

## Known limits of the validation

* **Performance was not measured** against the DPA-only baseline. Patch 4's vote adds one
  1-element host collective per `draft()` call; the cost is expected to be negligible but is
  unquantified.
* All runs used `--disable-custom-all-reduce` (required on gfx942/gfx950 for EAGLE), so the
  custom all-reduce path remains unexercised.
* Context 32768, short prompts, 512-token outputs, one image, one hardware configuration.
  Long-context and 400k-context configs are untested.
* The long durability run (1024/1024) was on the **mix** endpoint; the PD sweep was
  1384/1384 but shorter.
* A seventh occurrence of the padded-vs-real-rows crash family was seen **once** in 500+
  requests during an earlier session, on a machine already carrying patch 1. It did not recur
  in 2540 requests here, but at that rate absence is not evidence.

## Upstream

Patch 1 is ROCm-specific; the CUDA path was always correct.

Patch 4 is **not** ROCm-specific in principle. The guard it repairs is platform-neutral, and
its terms are rank-dependent on any backend running DP-attention with PD disaggregation; CUDA
is protected today only because the draft-extend graph is captured there, which keeps the
ranks on the same path. Upstream has related open work that blames `can_cuda_graph`
(#32209 all-gathers that decision, CI red; #32527 same topology; #32722 shows **no CI covers
PD+DPA+MTP**), but none of them addresses this guard — and `can_cuda_graph` was measured to
be *uniform* across ranks here, so #32209's vote would not have fixed it. This mechanism
appears unreported.

## Reproduction kits

Full environment capture, exact commands, raw logs, per-round evidence, root-cause notes and
the wrong turns taken live in the `infera.yihou.glm5.2.mxfp4` workspace:

* `glm52.mxfp4.spur.mooncake.packup_20260728/dpa_mtp_fix/` — patches 1–3
* `glm52.mxfp4.spur.mooncake.packup_20260729_bug2b_draft_graph/` — patch 4, with its
  differential control
* `glm52.mxfp4.spur.mooncake.packup_20260729_degenerate_output/` — falsification of an
  earlier "~2% degenerate output under concurrency" claim, which turned out to be a
  test-harness error (raw prompts posted to `/generate`, skipping GLM-5.2's chat template, and
  `temperature=0` forced over the model's recommended 1.0/0.95). With the template and the
  model's own sampling: 0/128 degenerate at conc=128 on both MXFP4 and FP8.


dorado269:
Superseded by #58 — closing.

This branch was cut from `cf85272`. `main` has since moved 29 commits ahead and absorbed three of the things carried here, so #58 is the same work rebased onto `8692fb4` with those dropped:

- `deepseek_nextn_glm52_mtp_bf16.diff` → `main`'s `patch_glm52_nextn_quark_exclude.py` (0d8d0ff) makes the identical edit. Keeping both would have been actively broken, not merely redundant: `main`'s patch loop runs first, so the context diff here would then fail at `--fuzz=0` against an already-edited anchor. #58 asserts it as a prerequisite instead.
- `build_mooncake_dmabuf.sh` → `main`'s `build_mooncake_sglang.sh` is a strict superset.
- `Dockerfile.sglang.dmabuf` → `main`'s `Dockerfile.sglang` now compiles dma-buf in and switches at runtime; the DSA layer moved into it, defaulting on.

Worth noting for anyone reading back through this PR: the rebase applied with **zero conflicts**. That was misleading rather than reassuring — this branch only *added* files, so git had nothing to compare against changes `main` made to the same problems in different files. The overlaps had to be found by reading, not by trusting the merge.

Net 8 files / 1181 lines → 6 files / 1048 lines. #58 re-validated the result end to end on the rebased branch (4/4 probe, 32/32, 128/128 ×2, 0 tracebacks) and also documents a configuration-only alternative to two of the patches (IndexShare off).

The branch `worktree-dsa-hip-dp-rows-fix` is left in place; nothing here is lost.

PR #19, opened 2026-07-23
feat(sglang): unified Mooncake image — one engine.so, every PD path runtime-decided

dorado269:
## Summary

Rebuild the v0.5.15.post1 base image's bundled Mooncake (**#2682**) **in place** so **one** sglang image supports **every** PD KV-transfer configuration, decided entirely at **runtime** — no per-feature image variant, no extra Mooncake version. Supersedes this PR's original HIP-gate-only scope, fixing its two gaps.

**Runtime knobs (launch-time, not build-time):**
- `MC_ENABLE_HIP_TRANSPORT` — HIP intra-node P2P transport, default **OFF** so cross-node PD stays on RDMA (#2682 installs HIP unconditionally and prefers it, breaking cross-node — `hipIpcOpenMemHandle` can't open a peer's handle).
- `MOONCAKE_DISABLE_HIP_DMABUF` — dma-buf GPUDirect (`ibv_reg_dmabuf_mr`) vs bare `ibv_reg_mr`+peermem, decided per memory-type/kernel by **#2682's own runtime switch** (already ships in the base, just compiled out — see below).
- single- vs multi-node — same image either way.

## What changed vs this PR's original approach

1. **Dedup.** The HIP-transport gate was a hand-written python string-replace, duplicating `deploy/docker/patches/mooncake_cpp/transfer_engine_impl.diff` that the vLLM build already applies. `build_mooncake_sglang.sh` now calls the **same** reusable `apply_mooncake_cpp_patches.sh` (HIP gate + auto-chunk MR); the duplicate `patch_mooncake_sglang.sh` is **deleted**.
2. **dma-buf now compiled in.** #2682 defines `USE_HIP_DMABUF` only on the `transfer_engine` target, but the `ibv_reg_dmabuf_mr` call lives in `rdma_context.cpp` → the `rdma_transport` OBJECT lib, so the branch was silently preprocessed out and the shipped `.so` used bare `ibv_reg_mr` (not even linking `libhsa-runtime64`). The build now propagates `USE_HIP_DMABUF` + `hsa-runtime64` to `rdma_transport` and builds `-DUSE_HIP_DMABUF=ON`.

[...]

PR #37, opened 2026-07-29
Enable PD-disaggregation e2e on dma-buf-only RoCE fabrics (crusoe)

limou102:
# Enable PD-disaggregation e2e on dma-buf-only RoCE fabrics (crusoe), for all 3 engines

## Summary

PD-disaggregated e2e could not run on the crusoe `amd-spur` cluster: every engine's
prefill worker died while registering its KV pool for RDMA. This enables it for
**vLLM, SGLang and ATOM**, verifies the KV really moves over GPU Direct RDMA, and
wires it into CI as a new `e2e-disag` job.

```
tests/run_tests.sh e2e disag          # all three engines, ~12 min
tests/run_tests.sh e2e vllm disag     # one engine
```

## Root cause

These nodes load no `ib_peer_mem`, so a bare `ibv_reg_mr()` on a device pointer
cannot pin VRAM. Measured directly with Mooncake's `TransferEngine`:

| image | dma-buf compiled in | `register_memory(VRAM)` |
|---|---|---|
| vLLM (before) | no | `rc=-202` FAIL |
| ATOM (before) | no (base image's old Mooncake) | `rc=-202` FAIL |
| SGLang | yes | `rc=0` OK |

`build_mooncake_rocm.sh` dropped the dma-buf registration path in #154 (it exhausts
a KFD resource at high util on ib_peer_mem fabrics → HIP-209). That is the right
default there, but on a dma-buf-only fabric it is the *only* path that works.

Two further findings:

- **Rail** — with dma-buf the ionic rail dies registering a production-sized KV pool
  (prefill worker killed during KV-cache init at `gpu-memory-utilization 0.9`). The
  flat mlx5 rail works; its routable GID is **index 3** (0 and 1 are link-local),
  unlike ionic's index 1.
- **ATOM silently fell back to TCP** — it names its own Mooncake NIC and guesses the
  MI300X `rdma<N>`. No such device here → `Found 0 HCAs` → TcpTransport. The case
  still *passed* all correctness probes; only the log showed it (and it ran 2.4x
  slower: 589s over TCP vs 247s over RDMA).

### Image build — dma-buf is opt-in, default behaviour unchanged

- Restore the B.1 CMake patch (`rdma_transport_dmabuf_cmake.diff`) that propagates
  `USE_HIP_DMABUF` to the `rdma_transport` object lib, where `rdma_context.cpp`
  actually calls `ibv_reg_dmabuf_mr`. Gated behind a new `MOONCAKE_HIP_DMABUF`
  build ARG, **default 0**.
- `Dockerfile.atom`: build Mooncake from the same pinned ref vLLM uses, but *only*
  when `MOONCAKE_HIP_DMABUF=1`; at the default the base image's Mooncake is
  untouched. Entrypoint switched to the host-libionic injector (see below).
- `build_mooncake_rocm.sh`: asserts the dma-buf symbols actually landed in the
  built `.so`, so a CMake define that fails to reach the target is a loud build
  failure rather than a silent fallback.

Verified equivalent at the default: a `MOONCAKE_HIP_DMABUF=0` build still reports
`bare ibv_reg_mr, needs ib_peer_mem`, applies B.2/B.3 only, and produces a `.so`
with **0** dma-buf symbols — byte-for-byte the pre-existing behaviour.

### Harness — fabric config is per-run, not per-engine

New cluster-level knobs, so no engine adapter or matrix row carries site specifics:

| env | purpose |
|---|---|
| `INFERA_E2E_BUILD_ARGS` | `K=V,K=V` → `docker build --build-arg` |
| `INFERA_E2E_WORKER_ENV` | `K=V,K=V` merged into every PD worker |
| `INFERA_E2E_GID_INDEX` | RoCE GID index (default 1) |
| `INFERA_E2E_SRUN_EXTRA` / `INFERA_E2E_RESERVATION` | extra `srun` flags |

Also:
- A build-arg'd image gets its **own tag** (`…-mooncake_hip_dmabuf1`), so the
  mixed/unit/engine tiers building the plain tag on the same nodes can't clobber it
  (they run as concurrent CI jobs).
- `cluster.srun_argv()` is now the single place scheduler flags are assembled;
  `launcher._srun` and the topology probes shared duplicated (and divergent) copies.
- `_has_libionic` probes **on the node** instead of on the orchestrator, which is a
  login box with no ionic stack — the mount was being dropped everywhere.
- ATOM's `ib_device` defaults to the rail the cluster pinned via `MC_TE_FILTERS`.
- Dropped the now-unused `shell_entrypoint` plumbing (all three images share one
  entrypoint contract).

### New: the test asserts RDMA, not just correctness

`assert_rdma_kv_transport()` fails a case whose worker logs show a TCP transport or
`Found 0 HCAs`. A TCP fallback still delivers the KV and still passes every
correctness probe, so nothing else distinguishes it. This is what caught ATOM.

```
[e2e disagg transport] prefill @ node-a: rdma
[e2e disagg transport] decode  @ node-b: rdma
```

### Scheduler robustness in `run_tests.sh`

- **QoS fallback.** Dispatches start on the cluster's default QoS; if a job sits in
  `QOSGrpNodeLimit` for 60s the watchdog cancels it and resubmits once with
  `--qos=amd-burst-qos`. Refused again → fail. Applies to `unit` / `engine` /
  `e2e mixed`; the disag tier does the equivalent with a one-shot probe before the
  stack comes up.
- **Queue visibility.** A queued job printed *nothing*, so a CI run looked hung and
  got cancelled. It now reports `still QUEUED on SLURM after Ns — job N, reason=…`
  every 60s.
- **Fixed a dead watchdog.** The job-id regex only matched `srun: job N running on …`,
  never `Pending job allocation N` — so the `JobHoldMaxRequeue` retry could never
  fire, since that state only exists while pending.
- Reservation-aware node picking (a reserved node reads `resv`, never `idle`, so
  `sinfo` alone returned nodes outside the pool), fallback to the open partition
  when the reservation has expired, and removal of leftover PD containers on
  Ctrl-C / CI cancel (pytest's fixture teardown does not run when killed).

### Test matrix

One case per engine: `gpt-oss-120b`, TP2 prefill + TP2 decode, Mooncake, production
memory utilisation (0.9). The vLLM `GLM-5.1-FP8` MoRIIO case and its adapter
plumbing are removed — see below.

### CI

New `e2e-disag` job (matrix: vllm, sglang, atom) with its own `run_e2e_disag`
dispatch input. The existing `run_e2e` input is renamed `run_e2e_mixed` for symmetry
— **update any `gh workflow run -f run_e2e=…` invocations.**

## Verification

All on crusoe `amd-spur`, 2 nodes, via `tests/run_tests.sh`:

| engine | case | result | time |
|---|---|---|---|
| sglang | gpt-oss-120b tp2 | PASS, `rdma` both sides | 294s |
| vllm | gpt-oss-120b tp2 | PASS, `rdma` both sides | 189s |
| atom | gpt-oss-120b tp2 | PASS, `rdma` both sides | 232s |

Both `e2e disag` (all three) and `e2e vllm disag` (single engine) end in
`RESULT: PASS`. Worker logs confirm `installTransport, type=rdma` on `mlx5_0` at
GID index 3. The QoS fallback was exercised against a real `QOSGrpNodeLimit`:
resubmission on `amd-burst-qos` started within seconds while default-QoS jobs stayed
queued.

## Backward compatibility

- `MOONCAKE_HIP_DMABUF` defaults to 0 — ib_peer_mem fabrics keep bare `ibv_reg_mr`,
  verified by rebuilding and diffing the resulting Mooncake `.so`.
- `run_e2e_mixed` / `run_unit` / `run_engine` are untouched; `pd_mixed` imports none
  of the modified harness modules.
- ATOM's entrypoint changed from `/bin/bash` to the host-libionic injector, which is
  required for RDMA (its container libionic is ABI-mismatched, so libibverbs
  enumerates 0 devices). The injector now execs bash when handed a leading flag, so
  `docker run  -c '…'` keeps working; `docker run  python3 …` now works
  too. Verified on all three images.
- The changed `COPY`/`RUN` lines invalidate the docker layer cache, so the first
  build after merge rebuilds the Mooncake layer once per node (~4-5 min).

## Known limitations

- **MoRIIO PD cannot run on this fabric.** `libmori_io` registers KV with bare
  `ibv_reg_mr` and ships no dma-buf path, so `RegisterRdmaMemoryRegion` returns
  `EINVAL` and the prefill worker dies mid-request. Pinning `MORI_RDMA_DEVICES`
  fixes an earlier `std::bad_cast` but not this. The GLM-5.1-FP8 MoRIIO case is
  therefore removed rather than skipped; restore it from history once MoRI-IO can
  register through dma-buf.
- The disag tier has no spill-to-open-partition path (unlike `_dispatch_slurm`); if
  the reservation is full it skips rather than borrowing nodes.
- On a build failure the disag retry excludes *both* nodes of the pair, which can
  exhaust a small reservation.

### DIFF
--- a/deploy/docker/patches/mooncake_cpp/apply_mooncake_cpp_patches.sh
+++ b/deploy/docker/patches/mooncake_cpp/apply_mooncake_cpp_patches.sh
@@ -38,16 +38,16 @@ apply_one() {
     fi
 }
 
-# B.1 (rdma_transport_CMakeLists.diff, USE_HIP_DMABUF -> ibv_reg_dmabuf_mr) REMOVED.
-# That dma-buf GPUDirect registration path exhausts a driver/KFD resource at high
-# gpu_memory_utilization (>=0.7): registering the whole KV pool (~156 GiB/GPU at
-# util 0.8) via ibv_reg_dmabuf_mr makes EVERY subsequent hipModuleLoad fail with
-# HIP-209 "no kernel image is available for execution", crashing the decode engine
-# at kernel_warmup / first inference (_compute_slot_mapping_kernel, torch fill_,
-# ...). The GPU BAR is 512 GiB (not the limit); reproduces on Kimi-K2.6 AND
-# DeepSeek-V4-Pro. Bare ibv_reg_mr (VRAM RDMA via host-libionic injection) does NOT
-# exhaust it and transfers KV correctly on ionic (validated util 0.8 P<->D, both
-# models). The diff is dropped entirely — bare ibv_reg_mr is the only supported path.
+# B.1 (USE_HIP_DMABUF -> ibv_reg_dmabuf_mr) is OPT-IN. It was dropped in #154
+# because dma-buf at high util exhausts a KFD resource (later hipModuleLoad ->
+# HIP-209) where ib_peer_mem is available; but on a dma-buf-only fabric (no
+# ib_peer_mem) it is the ONLY path that can register VRAM at all.
+if [ "${MOONCAKE_HIP_DMABUF:-0}" = "1" ]; then
+    echo "MOONCAKE_HIP_DMABUF=1 -> applying B.1 (dma-buf GPUDirect MR registration)"
+    apply_one rdma_transport_dmabuf_cmake.diff
+else
+    echo "MOONCAKE_HIP_DMABUF=0 (default) -> B.1 not applied (bare ibv_reg_mr; needs ib_peer_mem)"
+fi
 apply_one transfer_engine_impl.diff
 apply_one rdma_auto_chunk_mr_2017.diff
 
--- a/deploy/docker/patches/mooncake_cpp/rdma_transport_dmabuf_cmake.diff
+++ b/deploy/docker/patches/mooncake_cpp/rdma_transport_dmabuf_cmake.diff
@@ -1,0 +1,39 @@
+diff --git a/mooncake-transfer-engine/src/transport/rdma_transport/CMakeLists.txt b/mooncake-transfer-engine/src/transport/rdma_transport/CMakeLists.txt
+--- a/mooncake-transfer-engine/src/transport/rdma_transport/CMakeLists.txt
++++ b/mooncake-transfer-engine/src/transport/rdma_transport/CMakeLists.txt
+@@ -7,3 +7,35 @@ target_link_libraries(rdma_transport PRIVATE JsonCpp::JsonCpp glog::glog
+ if(USE_MLX5DV)
+   target_compile_definitions(rdma_transport PRIVATE USE_MLX5DV)
+ endif()
++
++# infera B.1: enable the HIP dma-buf GPU-direct MR registration path for THIS
++# object library. rdma_context.cpp (the file that actually calls
++# ibv_reg_dmabuf_mr / hsa_amd_portable_export_dmabuf) is compiled into the
++# `rdma_transport` OBJECT lib, NOT into `transfer_engine`. Upstream only defines
++# USE_HIP_DMABUF on the `transfer_engine` target (src/CMakeLists.txt), so the
++# #elif defined(USE_HIP_DMABUF) branch in rdma_context.cpp compiles OUT here and
++# GPU buffers fall through to the bare ibv_reg_mr() branch. Bare ibv_reg_mr on a
++# device pointer only works when the legacy ib_peer_mem kernel module is loaded;
++# on a dma-buf-only RoCE fabric (no ib_peer_mem) it cannot pin VRAM and the KV
++# registration fails (huge host-memory pin / OOM). Mirror the transfer_engine
++# block so the dmabuf verb is compiled into the code path that uses it.
++#
++# USE_HIP is a global compile definition (mooncake-common/common.cmake) and the
++# option is a cache var, so both are visible here. hsa/hip headers are already on
++# the global include path (include_directories(${HIP_INCLUDE_DIRS})); we still
++# link hsa-runtime64 so hsa_amd_portable_export_dmabuf resolves. At runtime
++# rdma_context.cpp::isKernelDmabufSupported() self-checks CONFIG_PCI_P2PDMA /
++# CONFIG_DMABUF_MOVE_NOTIFY and falls back to ibv_reg_mr if the kernel lacks
++# them, and MOONCAKE_DISABLE_HIP_DMABUF=1 forces the old path — so this is safe
++# to compile in unconditionally on the images that opt into it.
++if(USE_HIP)
++  option(USE_HIP_DMABUF "Enable HIP dmabuf RDMA MR registration" ON)
++  if(USE_HIP_DMABUF)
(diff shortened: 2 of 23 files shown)


deploy/docker/patches/mooncake_cpp/rdma_transport_dmabuf_cmake.diff
```diff
@@ -0,0 +1,39 @@
...
++#
++# USE_HIP is a global compile definition (mooncake-common/common.cmake) and the
++# option is a cache var, so both are visible here. hsa/hip headers are already on
++# the global include path (include_directories(${HIP_INCLUDE_DIRS})); we still
++# link hsa-runtime64 so hsa_amd_portable_export_dmabuf resolves. At runtime
++# rdma_context.cpp::isKernelDmabufSupported() self-checks CONFIG_PCI_P2PDMA /
++# CONFIG_DMABUF_MOVE_NOTIFY and falls back to ibv_reg_mr if the kernel lacks
++# them, and MOONCAKE_DISABLE_HIP_DMABUF=1 forces the old path — so this is safe
++# to compile in unconditionally on the images that opt into it.
++if(USE_HIP)
```
dorado269, line 29: the latest mooncake already added USE_HIP_DMABUF by default and fix USE_HIP but missing dmabuf obj?

deploy/docker/patches/mooncake_cpp/apply_mooncake_cpp_patches.sh
```diff
@@ -38,16 +38,16 @@ apply_one() {
...
-# That dma-buf GPUDirect registration path exhausts a driver/KFD resource at high
-# gpu_memory_utilization (>=0.7): registering the whole KV pool (~156 GiB/GPU at
-# util 0.8) via ibv_reg_dmabuf_mr makes EVERY subsequent hipModuleLoad fail with
-# HIP-209 "no kernel image is available for execution", crashing the decode engine
-# at kernel_warmup / first inference (_compute_slot_mapping_kernel, torch fill_,
-# ...). The GPU BAR is 512 GiB (not the limit); reproduces on Kimi-K2.6 AND
-# DeepSeek-V4-Pro. Bare ibv_reg_mr (VRAM RDMA via host-libionic injection) does NOT
-# exhaust it and transfers KV correctly on ionic (validated util 0.8 P<->D, both
-# models). The diff is dropped entirely — bare ibv_reg_mr is the only supported path.
+# B.1 (USE_HIP_DMABUF -> ibv_reg_dmabuf_mr) is OPT-IN. It was dropped in #154
```
dorado269, line 41: 1. You can always compile the DMABUF in and disable it using some flag when running
2. dma-buf way acts correct when the net card has ODP support.

some work is done for sglang engine, see: https://github.com/AMD-AGI/Infera/pull/19

dorado269: https://github.com/AMD-AGI/Infera/pull/33
and this pr add a tool to detect the rdma env, then recommand the best choice and provide the flag

PR #55, opened 2026-07-31
feat(k8s): Kubernetes recipes — single-node (mixed) + PD 1P1D (Mooncake)

jiejingzhangamd:
Copy-paste, **locally-validated** `InferaDeployment` recipes for running Infera on Kubernetes via the operator — the Infera analog of a vendor "recipe" — plus the engine-image fix they needed.

## What's here

**Mixed (single-node aggregated)** — router + one mixed worker, validated end to end on k3s (MI355X) to real tokens:
- `single-node-qwen-sglang.yaml` / `single-node-qwen-vllm.yaml` + `RECIPE-single-node.md` (the sglang↔vllm diff is *not* just the image)
- `model-cache/` — a PVC + download Job (the Dynamo model-cache analog); validated Job → PVC → serve
- `kimi-k3-vllm.yaml` + `manual/examples/k8s_kimi_k3.md` — Kimi-K3 TP8 multimodal, full operator → router → `infera.engine.vllm` path (built infera on the `kimi_k3` vLLM base; needs `--kv-cache-dtype auto` so MLA doesn't pick the batch-1-only `mla_gluon` kernel)

**PD-disaggregated 1P1D (Mooncake)** — router + prefill (node A) + decode (node B), KV over the ionic RoCE fabric:
- `pd-1p1d-mooncake.yaml` + `manual/examples/k8s_pd_1p1d_mooncake.md`
- Validated on two-node k3s through operator reconcile, cross-node registration, PD dual-dispatch, in-pod RDMA discovery, and the pod-overlay Mooncake RPC. The final KV move needs the two nodes on a **mutually routable RoCE fabric** (documented).

**Engine image libionic fix**
- `Dockerfile.sglang` + `Dockerfile.sglang.gfx942`: bake ABI-4 libionic (`INSTALL_LIBIONIC=1`). The base ships ABI-1 libionic, so `ibv_get_device_list` returns 0 HCAs and Mooncake silently drops to TCP → cross-node PD KV transfer fails. With the fix `ibv_devinfo` sees the 8 ionic HCAs in-pod without the runtime host-lib injection.

## Validation
Single/two-node k3s (MI355X, `amd.com/gpu`): sglang + vllm mixed serving, Kimi-K3 TP8 multimodal, model-cache Job→PVC→serve, and PD reconcile/register/dual-dispatch up to the cross-node RoCE-fabric boundary.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #60, opened 2026-07-31
docs: field notes for working on the AMD ROCm inference stack

jiejingzhangamd:
A handbook for whoever works on this stack next — human or agent. Consolidates what was learned bringing up Kimi-K3 and GLM-5.2 on MI355X, building the Kubernetes recipes (#55), and porting a C++ inference engine to ROCm.

Everything in it cost real debugging time, and none of it is guessable from reading the source.

**Organising principle: on this stack the expensive failures are silent.** Things do not crash — they degrade, or return plausible numbers that are wrong. Every section is a case of that.

## Contents

**Silent failures, ranked by cost**
- libionic ABI mismatch → `ibv_get_device_list` returns 0 HCAs → RDMA falls back to TCP with no error. Cross-node PD "works" and is 15-50x slower. Includes the one-line check that actually detects it.
- Dead TP worker processes keep VRAM after `docker rm -f` / `kubectl delete`, so the *next* deployment fails with what looks like a config error. Cost two failed deployments before it was recognised.
- HIPIFY leaves warp-32 lane math untouched on a 64-lane wavefront: compiles cleanly, wrong answers, no diagnostic.

**Per-model traps** — these are model properties, not preferences, and copying flags between models is how you hit them. Kimi-K3 needs `--kv-cache-dtype auto` (fp8 KV selects the batch-1-only `mla_gluon` kernel), and cannot use prefix caching (hybrid Mamba) or `fastsafetensors` (needs GDS). GLM-5.2 decode is numerically broken on vLLM — use SGLang, with the ROCm DSA top-k env or it crashes at init.

**aiter as a C++ operator library** — the three entry layers and which to use, why to compile kernel TUs from source instead of linking the prebuilt `.so`s (those need libtorch), and that kernel availability is per-arch: gfx942 ships 58 asm blobs, gfx950 ships zero `pa_*.co`.

**Kubernetes** — k3s `--data-dir` on a large disk or DiskPressure evicts the operator; the node token follows the data-dir; containerd is not docker; model mounts need `extraPodSpec`; `startupProbe` over `skipReadinessProbe`; and cross-node RDMA needs a *routable* fabric, not merely active NICs.

**Method** — run it rather than reason about it (three wrong conclusions in this work were caught only by executing), verify a test can actually fail, and cross-check comments against runtime logic. One stale comment (`CachePoolMode`'s documented default is the opposite of what the code selects) would have led to the wrong port strategy.

### DIFF
--- a/docs/ROCM_AGENT_HANDBOOK.md
+++ b/docs/ROCM_AGENT_HANDBOOK.md
@@ -1,0 +1,252 @@
+# Working on the AMD ROCm inference stack — field notes
+
+Notes for whoever (or whatever) works on this stack next. Everything here cost
+real debugging time; none of it is guessable from the source alone. Hardware is
+MI300X (`gfx942`) and MI355X (`gfx950`), ROCm 7.2.
+
+The organising principle: **on this stack, the expensive failures are silent.**
+Things do not crash — they degrade, or return plausible numbers that are wrong.
+Every section below is a case of that.
+
+---
+
+## 1. Silent failures, ranked by how much time they cost
+
+### RDMA quietly becomes TCP (libionic ABI)
+
+Engine images ship libionic built against an old kernel ABI (e.g. `54.0-149` =
+ABI 1). The MI355X ionic driver wants ABI 4 (`54.0-187`). On mismatch
+`ibv_get_device_list` returns **0 HCAs**, and Mooncake/MoRI fall back to TCP
+without an error. Cross-node PD "works" and is 15-50x slower, or fails later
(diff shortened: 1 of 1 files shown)


jiejingzhangamd:
Moved to the dash-infer repo instead — it sits next to `docs/AMD_ROCM_PORT.md` there, which is the plan these notes support, and next to the `tools/rocm_poc/` PoCs they refer to. Now `docs/ROCM_FIELD_NOTES.md` on kzjeef/dash-infer#691.

PR #81, opened 2026-08-03
docs(recipes): correct the port-collision note — the fix landed on main

jiejingzhangamd:
The GLM-5.2 header recorded "No `Address already in use`" from the DP-attention run, worded so it reads as *"main has this bug and we did not trip it"*. Two things are wrong with that now.

**The run predates #79**, which landed the randomised scan start in `free_tcp_port_block` on main — the same fix #59 carried. The risk is retired, not merely untriggered.

**And the observation proved less than the wording implied even at the time.** That collision needs two engines on **one** host; a cross-node PD pair cannot produce it. Recording a clean result from a topology that cannot fail the test reads as evidence when it is not.

Both the manifest header and the README table now say so.


jiejingzhangamd:
Folded into #80 so the 0.2.0 prep is one review. Same commit, cherry-picked.

PR #84, opened 2026-08-04
docs(examples): add a GLM-5.2 1P1D SGLang deployment example

dorado269:
# Description

Adds a runnable deployment example for **GLM-5.2-MXFP4** under `examples/sglang_1p1d_glm5.2/`:
one prefill node and one decode node, KV moved between them over Mooncake RDMA, fronted by the
infera router, with DP-attention, MTP (EAGLE speculative decoding) and the kvd cache tiers on.

Everything site-specific lives in one of two files under `cluster/` — one per RDMA registration
mode, picked by the preflight probe rather than by autodetection at launch. `common.sh` and
`engine/*.sh` carry the tuned recipe and contain no addresses, paths, NIC names or GID indices,
so adapting to a cluster means editing a wrapper, never an engine script.

Directory shape mirrors `examples/deepseek_v4` (`common.sh` + an `engine/` subdirectory).

## Type of change

- [x] Documentation change (change only to the documentation, either a fix or a new content)
- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- `cluster/` — two wrappers, the only files a user edits: `cluster.peermem.sh` (multi-rail
  with a peer-memory module) and `cluster.dmabuf.sh` (single ODP NIC, no peer-mem). No
  autodetection: `preflight_rdma.sh mode` tells you which one to use.
- `engine/leg.sh` — the launcher for one PD leg. Carries the tuned recipe and nothing
  site-specific.
- `engine/up.sh` / `down.sh` — bring up both nodes (containers → etcd + kvd → both legs →
  router) and tear down, waiting for VRAM to actually drain.
- `engine/smoke.sh` — service check **plus** positive evidence for each of the five features.
  A green `/health` proves the process is alive, not that PD paired or that speculation is
  doing anything; every check here goes red if its feature is silently absent.
- `engine/bench.sh` — reference throughput sweep using SGLang's own `bench_serving`.
- `preflight_rdma.sh` — registration-mode probe and cross-node fabric measurement.
- `results/` — concurrency-8 numbers from two independent agentic benchmarks on two clusters.

### Three couplings that fail quietly

Each of these produces a plausible result rather than an error when got wrong, so each is
commented at the point of use and written up in the README's "Notes & gotchas":

- **`--ep-size` is emitted unconditionally**, outside the DP-attention branch. Expert and
  attention parallelism are different axes; gating both on one condition collapses the MoE
  from ep8 to the TP default whenever DP-attention is off, so a run billed as "DPA off"
  differs in the expert-dispatch collective too and no latency delta is attributable.
- **`--chunked-prefill-size` is a global budget** that SGLang divides by `dp_size` only when
  DP-attention is on. One value serves both modes; hardcoding the per-rank number in a DPA-off
  branch cuts the global budget 8×.
- **Prefill activation OOM is fixed by *lowering* `--mem-fraction-static`** — the opposite of
  the decode-side retract fix. Low token usage at the moment of the abort is the tell: the KV
  pool is nearly empty, so it was never KV exhaustion.

### On `results/`

Numbers are recomputed from raw per-request records, not copied from summary lines. The
cross-cluster comparison is presented as **context, not a measurement**: more than one variable
moves between those clusters, four candidate causes are listed with the evidence behind each,
and none is confirmed. Where a benchmark's own summary metric is misleading it is called out —
the customer harness's `Theoretical Prefix Cache Hit` is computed from the trace file and never
asks the server, so it is invariant to the deployment under test.

The customer's benchmark is **referenced by URL, not vendored**; this kit ships no agentic bench
client, only the service self-check and the `bench_serving` reference sweep.

# Checklist:

**No tests.** This is a documentation/example change: shell scripts that drive a two-node GPU
deployment, with no unit-testable surface and no CI cluster to run them against. What was done
instead is recorded below.

## Validation status

Stated plainly rather than implied — the README carries the same table.

| what | status |
|---|---|
| the deployment shape (1P1D + mooncake + DPA + MTP + kvd + kv-aware) | **validated end-to-end on two clusters**, both fabric types |
| the tuned values (GMU, chunk, ctx, EAGLE settings, DSA env, router weights) | **validated** — each carried over from a run that completed cleanly |
| **these scripts as written** | **validated** — `preflight mode` → `up` → `smoke` → `bench` → `down` on a 2-node MI355X mode-B cluster, no edits outside the wrapper. Long context checked separately (needle, to 238K tokens) and under a real agentic workload at concurrency 8 |
| `preflight_rdma.sh fabric` | **not validated** — only `mode` was exercised |
| `cluster.peermem.sh`, `round-robin` routing | **not validated** — no peer-mem cluster was available; the shipped `kv-aware` default is what ran |

Static checks: `bash -n` on all 9 scripts; every infera flag grepped against `infera/**/args.py`
on `main`; the deliverable grepped for leaked local information (host names, absolute paths,
scheduler identifiers, credentials) before submission.

## Known placeholder

The image tag `inferaimage/infera-sglang:0.2.0` is a **placeholder**, flagged as such in the
README. It needs replacing with the released infera-sglang tag before this is useful to a
reader.

### DIFF
--- a/examples/sglang_1p1d_glm5.2/cluster/README.md
+++ b/examples/sglang_1p1d_glm5.2/cluster/README.md
@@ -1,0 +1,140 @@
+# Adapt to your cluster — change things HERE
+
+Everything site-specific in this kit lives in one of the two wrapper files in this
+directory. `common.sh` and `engine/*.sh` carry the tuned recipe and contain no
+addresses, paths, NIC names or GID indices at all. **If you need to adapt the
+deployment to your cluster, edit a wrapper — not the engine scripts.**
+
+## Which wrapper?
+
+Run the mooncake-mode probe on one of your nodes first. It reads the node, not a
+config file, and prints which registration modes are viable:
+
+```bash
+bash preflight_rdma.sh mode
+```
+
+Then pick by what it says about **peer-mem**:
+
+| preflight says | wrapper | why |
+|---|---|---|
+| `peermem: present`, **mode A viable** | [`cluster.peermem.sh`](cluster.peermem.sh) | bare `ibv_reg_mr` hands the NIC the GPU pages directly — nothing pinned, KV pool not duplicated, **every** rail can carry KV |
+| `peermem: absent`, **mode B viable** (an ODP NIC exists) | [`cluster.dmabuf.sh`](cluster.dmabuf.sh) | dma-buf is the only GPUDirect path without peer-mem, and it is only safe on an ODP NIC; KV is locked to that one NIC |
+| **only mode C viable**, or nothing viable | *neither, yet* | mode C pins and **doubles** the KV pool and needs an explicit, model/TP/VRAM-specific cap. Preflight exits `2` in this case on purpose: it is a human decision, not a default. |
+
+The probe does not just classify — for the mode it picks it prints the exact env
+and launch flags. The wrapper fields below are a place to paste those values, not
+a puzzle to solve.
+
+## The fields you must fill in
+
+Both wrappers have the same four blocks. Placeholders read `<like-this>`.
+
+### 1. Nodes
+
+| var | what |
+|---|---|
+| `PREFILL_NODE` / `DECODE_NODE` | names the bring-up host can run a command on |
+| `PREFILL_IP` / `DECODE_IP` | each node's **data-plane** IP — the address on the KV network |
+| `KIT_DIR` | where this kit lives; must be the **same path on both nodes** |
+
+**The data-plane IP is not the management IP.** The legs advertise these addresses
+to each other and to the router for the KV handoff. A worker that advertises a
+non-routable address (`0.0.0.0`, `127.0.0.1`) is rejected at launch by infera's own
+config preflight, but a *wrong-but-routable* one is not — it just hangs.
+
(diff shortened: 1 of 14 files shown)


examples/sglang_1p1d_glm5.2/cluster/README.md
```diff
@@ -0,0 +1,140 @@
...
+```bash
+bash preflight_rdma.sh mode
+```
+
+Then pick by what it says about **peer-mem**:
+
+| preflight says | wrapper | why |
+|---|---|---|
+| `peermem: present`, **mode A viable** | [`cluster.peermem.sh`](cluster.peermem.sh) | bare `ibv_reg_mr` hands the NIC the GPU pages directly — nothing pinned, KV pool not duplicated, **every** rail can carry KV |
+| `peermem: absent`, **mode B viable** (an ODP NIC exists) | [`cluster.dmabuf.sh`](cluster.dmabuf.sh) | dma-buf is the only GPUDirect path without peer-mem, and it is only safe on an ODP NIC; KV is locked to that one NIC |
```
dorado269, line 22: verified

xiaobochen-amd (approved):

PR #85, opened 2026-08-04
fix(sglang-rocm): make kvd hicache work on the gfx942 v0.5.16 base — one crash, two wrong diagnostics

llying-001:
# Description

Three fixes that stand between a GLM-5.2-FP8 PD deployment on gfx942 and a working
kvd L3 tier. The load-bearing one is a scheduler crash; the other two are
diagnostics that told operators something untrue.

**Enabling kvd on the v0.5.16 gfx942 base crashes on the first reused prefix.**
Startup, weight load and a single-shot request all succeed, then the prefill
scheduler dies with exit -3:

```
tvm.error.InternalError: Tensor match failed for
  Tensor<1152>[strides=<1>, dtype=int64, device=rocm:0]
- Root cause: Device value [rocm:0] not in the allowed options: [cpu, rocm_host]
```

Two gates for one decision disagree on ROCm. `pool_host/mla.py` opts HIP into the
staged write-back JIT kernel, while `DSAIndexerPoolHost` — the sidecar in the same
`HostPoolGroup` for any DSA model — still gates it on `_is_cuda`. The group ANDs
the flag over its entries, reads False, and moves the destination indices to the
GPU; it then delegates the write-back to its anchor, the MLA pool, which reads its
own flag, True, and launches the JIT kernel that needs those indices on the host.
The new patch gates MLA like every other host pool, so both sides read False and
the non-JIT kernel in the same branch takes over — which is what ROCm has always
run. It costs nothing that worked before: the group flag was already False here
regardless, so the `_is_hip` bought no faster path, it only desynchronized the two
gates.

Not specific to kvd — any hierarchical cache on this path dies the moment it writes
a page back. Upstream `main` is affected too. The proper repair is upstream's own
([sglang#30350](https://github.com/sgl-project/sglang/pull/30350), OPEN, stalled
since 07-16), which teaches the other pools the JIT instead; our patch only has to
stop the crash, and it refuses to apply once #30350 lands.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Infra/Build change
- [ ] Code refactoring

## Changes

**`fix(sglang-rocm): make hicache survive its first write-back on gfx942`**

- New `patches/sglang_rocm/patch_hicache_rocm_staged_write_back.py`: gates
  `can_use_write_back_jit` on CUDA in `pool_host/mla.py`. The anchor swallows
  upstream's comment along with the assignment, since that comment argues for the
  HIP enablement being removed — leaving it would contradict the patched code.
- Dropping the patch **cannot** be keyed on anchor drift: #30350 repairs the other
  pools and never touches `mla.py`, so the anchor would keep matching and the patch
  would keep gating the anchor back down, silently forfeiting the staged kernel
  #30350 enables. `check_group_still_poisoned()` checks that *precondition* instead
  and exits 1 once `DSAIndexerPoolHost` stops gating on `_is_cuda` alone.
- `Dockerfile.sglang.gfx942` now runs the shared `sglang_rocm/` loop, which also
  brings the existing hicache host-allocator patch to this image. That one is
  **preventive** there: MI300X was measured to return host VA == device VA from
  `hipHostRegister`, but that identity is a property of the driver rather than of
  the API, gfx942 is `xnack-` too, and the failure would land on the write-back
  path where it is expensive to diagnose.
- `Dockerfile.sglang` (mi35x, v0.5.15.post1) documents why the same script is a
  no-op there: every write-back gate on that base is still `_is_cuda`, so the group
  AND and its anchor agree. #28534 introduced the disagreement after that tag.
- `patch.upstream.status.md`: a row for the new patch, notes grouped per patch, and
  a `Maintenance` caveat — the new `MERGED` PR (#28534) is what *introduced* the
  defect, so the usual "upstream merged it, drop the row" rule does not apply.

**`fix(kvd): report the io_mode the L3 region actually resolved`**

- The startup self-check re-classified the mount instead of being told what the
  long region had settled on, so its GB/s line could describe a mode kvd is not
  using — either because `--io-mode` / `INFERA_KVD_IO_MODE` was explicit, or
  because the device walk dead-ends (in a container it stops at a `/dev/mapper`
  node outside the namespace).
- The file-per-block region now reports `False` outright, since it only ever opens
  buffered while the probe answers `direct` on NVMe. `None` still means "let the
  self-check classify" — the honest answer for `--tablespace-auto-detect-fs`, where
  the region resolves `o_direct` at `start()`.
- When caller and probe disagree, the self-check says so in the rationale it logs,
  so `io_mode=direct` no longer sits next to the probe's reason for picking
  buffered.

**`fix(sglang): correct two hicache diagnostics that misread current sglang`**

- The hicache-ratio guard attributed `prefetch_capacity_limit = 0.8 * (host_pool -
  device_pool)` to every sglang up to 0.5.15. That formula was replaced **in
  0.5.15** by `int(0.5 * host_pool)` — read from the raw `cache_controller.py` at
  v0.5.4 / 0.5.12 / 0.5.14 / 0.5.15.post1 / 0.5.16 / `main`. So the boundary is
  0.5.14, and above it a low ratio is a hit-rate loss (an L2 that evicts as fast as
  the L1 it shadows) rather than prefetch being switched off. The threshold constant
  does not move and the guard still fires on both; only the explanation an operator
  reads was wrong.
- `kvd_wiring` logged a WARNING when `ServerArgs` carried no `prefetch_threshold`
  field, telling the operator to pass an override that `_append_sglang_hicache_argv`
  already puts on the child argv a few lines later — and that argv copy is the one
  the engine subprocess reads. Demoted to INFO, and it now says where the 64 comes
  from.

## Verification

Both patch scripts were exercised against the **v0.5.16 and current-`main`** copies
of `pool_host/mla.py` + `memory_pool_host.py`, on throwaway trees:

| scenario | expected | result |
|---|---|---|
| stock v0.5.16 / `main` | apply, exit 0 | applies, patched file compiles, marker reaches the bytecode |
| re-run on a patched tree | exit 0 | `already applied` |
| #30350 simulated (three CUDA-only gates flipped) | refuse, exit 1 | refuses, `mla.py` byte-identical to pristine |
| no `pool_host/mla.py` (v0.5.15.post1 shape) | exit 0 | tolerated |
| anchor drifted / pool renamed / `memory_pool_host.py` absent | exit 1 | exit 1 with the offending lines printed |
| stale `__pycache__` | only `mla.*` / `common.*` removed | `mha.*` left in place |

kvd: a real daemon on a file-per-block long region now logs `io_mode=buffered`; a
caller/probe conflict logs `io_mode from the caller, not the probe (probe wanted
buffered: ...)`, and agreement adds no noise.

`ruff check` and `ruff format --check` clean. `tests/engine/sglang` +
`tests/unit/kvd`: 596 passed, 3 skipped. `tests/engine/vllm/test_mla_tp_dedup.py`
has 7 pre-existing failures on this machine (`PermissionError` on
`/nonexistent/model-for-test`) — reproduced identically on the untouched tree.

# Checklist:

- [x] New and existing unit tests pass locally with my changes

### DIFF
--- a/infera/kvd/storage_selfcheck.py
+++ b/infera/kvd/storage_selfcheck.py
@@ -103,6 +103,11 @@ def run_storage_selfcheck(
         d = Path(long_path) / ".kvd_selfcheck"
         d.mkdir(parents=True, exist_ok=True)
         r_od, r_workers, r_chunk, why = _resolve(Path(long_path))
+        if o_direct is not None and bool(o_direct) != r_od:
+            # Otherwise the log line reads `io_mode=direct` next to the probe's
+            # reason for picking buffered, and looks self-contradictory.
+            probed = "direct" if r_od else "buffered"
+            why = f"io_mode from the caller, not the probe (probe wanted {probed}: {why})"
         o_direct = r_od if o_direct is None else bool(o_direct)
         chunk = r_chunk if chunk_bytes is None else int(chunk_bytes)
         w_workers = int(write_workers) if write_workers else r_workers
--- a/deploy/docker/Dockerfile.sglang
+++ b/deploy/docker/Dockerfile.sglang
@@ -146,7 +146,7 @@ RUN set -eu; \
     for f in /tmp/sglang-disagg-patches/*.py; do echo "[sglang-patch] $f"; python "$f"; done; \
     rm -rf /tmp/sglang-disagg-patches
 
-# ---- sglang ROCm patches (hierarchical-cache host allocator) ----------------
+# ---- sglang ROCm patches (hicache host allocator, staged write-back) --------
 # REQUIRED FOR CORRECTNESS on ROCm whenever hierarchical cache / kvd is enabled:
 # hipHostRegister maps host pages at a device VA that differs from the host VA,
 # but sglang's hicache stores raw host data_ptr()s in device-side pointer tables
@@ -163,6 +163,14 @@ RUN set -eu; \
 # kvd writes back is worse than a failed build. Self-locating, idempotent, and
 # it plants a real module-level string literal so the fix can be proven in the
 # BYTECODE rather than in the source.
+#
+# The directory also carries patch_hicache_rocm_staged_write_back.py, which
+# resyncs two gates that disagree about the staged write-back JIT on ROCm. This
+# base is not affected: every gate here is still _is_cuda, so they agree on False
+# and the non-JIT kernel runs — sglang#28534 introduced the disagreement after
+# this tag. That script exits 0 on the absent mem_cache/pool_host/mla.py. Kept in
+# the shared directory so a base bump picks the fix up rather than rediscovering
+# it as a scheduler crash.
 COPY deploy/docker/patches/sglang_rocm/ /tmp/sglang-rocm-patches/
 RUN set -eu; \
     for f in /tmp/sglang-rocm-patches/*.py; do echo "[sglang-patch] $f"; python "$f"; done; \
--- a/deploy/docker/Dockerfile.sglang.gfx942
+++ b/deploy/docker/Dockerfile.sglang.gfx942
@@ -120,6 +120,42 @@ RUN set -eu; \
     for f in /tmp/sglang-disagg-patches/*.py; do echo "[sglang-patch] $f"; python "$f"; done; \
     rm -rf /tmp/sglang-disagg-patches
 
+# ---- sglang ROCm patches (hicache host allocator, staged write-back) --------
+# Two independent hicache-on-ROCm fixes, both needed once kvd is enabled.
+#
+# (1) host allocator. hicache allocates its host pools with mmap +
+# hipHostRegister and then hands the raw host data_ptr()s to GPU kernels through
+# device-side pointer tables, which is only correct while hipHostRegister happens
+# to map the pages at the host VA. The patch moves ROCm to hipHostMalloc, whose
+# device pointer IS the host pointer by contract.
+#
+# PREVENTIVE ON THIS BASE, not a fix for a crash seen here. On gfx950 the two
+# addresses differ and the first kvd write-back aborts the process with "Memory
+# access fault by GPU node-N"; on gfx942 they were measured equal (MI300X, amdgpu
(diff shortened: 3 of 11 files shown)


infera/kvd/storage_selfcheck.py
```diff
@@ -103,6 +103,11 @@ def run_storage_selfcheck(
         d = Path(long_path) / ".kvd_selfcheck"
         d.mkdir(parents=True, exist_ok=True)
         r_od, r_workers, r_chunk, why = _resolve(Path(long_path))
+        if o_direct is not None and bool(o_direct) != r_od:
+            # Otherwise the log line reads `io_mode=direct` next to the probe's
+            # reason for picking buffered, and looks self-contradictory.
+            probed = "direct" if r_od else "buffered"
+            why = f"io_mode from the caller, not the probe (probe wanted {probed}: {why})"
```
jiejingzhangamd, line 110: why we chnage this part for hicache?

llying-001: Reporting only — no I/O behaviour changes. The self-check was re-classifying the
mount instead of using the io_mode the L3 region actually resolved, so its GB/s
line could be labelled with a mode kvd isn't running.
We hit that on this cluster because our L3 is xfs on an **LVM** over 8 NVMe,
while `_lsblk_for_source()` was written against **md** — its docstring says so
(`e.g. /dev/md0`). Before the `--inverse` walk it rebuilds the child path as
`f"/dev/{name}"`, which holds for md (`md1` → `/dev/md1` exists) but not for
LVM: lsblk reports `nvme_vg-nvme_lv`, whose real node is `/dev/dm-8` and whose
only resolvable name is `/dev/mapper/nvme_vg-nvme_lv`. The walk dead-ends, the
probe answers "unknown device → buffered", and meanwhile the region is running
O_DIRECT from an explicit `--io-mode direct`. The self-check then printed a
14.51 GB/s O_DIRECT number next to the probe's reason for choosing buffered.
The walk bug itself is a separate fix and is not in this PR (it also fails one
step earlier inside a container, where `privileged: true` doesn't help for LVM —
the friendly name is a udev symlink and containers don't run udev). Happy to
split this commit out too if you'd rather keep the PR strictly to the SGLang
patches.

jiejingzhangamd (approved):

PR #96, opened 2026-08-06
docs(examples): GLM-5.2-FP8 PD on gfx942 — the docker recipe and its Kubernetes translation

llying-001:
# Description

GLM-5.2-FP8 was brought up, tuned and benchmarked on 2 × MI300X with SGLang
prefill/decode disaggregation, and nothing in the repo carried that deployment.
Anyone repeating it had to re-derive it, and the parts easiest to get wrong are
exactly the parts that fail quietly — a KV transfer that falls back to TCP, an L3
tier that is written and never read, kv-aware routing degraded to load balancing.

This lands the deployment twice: as the `docker` + shell scripts it was actually
validated with, and as the Kubernetes manifest that translates them. Every engine
and kvd flag in the manifest is the docker recipe's; section 6 of the recipe README
enumerates each place the substrate forced a difference, so a reader can check the
claim rather than trust it.

The values are the substance of this PR. They are not SGLang defaults, and each one
carries the measurement that chose it, taken one axis at a time on an agentic
multi-turn trace (32 conversations / 225 turns, CONC=16, ~68k-token median input):

| Setting | Value | What it bought |
| --- | --- | --- |
| `ROUTER_BACKEND` | `rust` | same routing decisions as `python`, 27% faster end to end |
| `CHUNK` | `8192` (1,024/rank) | −23.8% duration, −34.4% TTFT against 16,384/rank |
| MTP steps/topk/draft | `5`/`1`/`6` | −8.5% duration, −13.8% TPOT; 4.64 acceptance on a 4.00 break-even |
| `IB_DEVICE` | one rail | striping over every NIC measured 11.9% *slower* |
| dp-attention | both legs | pure TP8 prefill measured 25.9% slower |
| `KVD` | `0` | the offload tier cost 12% and served zero reads on this trace |

Together: 764 s → 420 s (−45%), output throughput 64.8 → 118.0 tok/s. One metric
moved the wrong way — ITL p90 rose 13.7%, because a deeper draft emits tokens in
burstier groups.

## Type of change

- [x] Documentation change (change only to the documentation, either a fix or a new content)
- [ ] Infra/Build change
- [ ] Code refactoring

Nothing outside `examples/` changes.

## Changes

- **`examples/recipes/glm5.2-fp8-gfx942/` — new Kubernetes recipe.** The deployment
  as one `InferaDeployment`: router + prefill (with a kvd sidecar) + decode, TP8/DP8
  with DP-attention, MTP, fp8 KV, Mooncake RDMA between the legs, KV offload to host
  RAM and node-local NVMe. kvd runs as a *native* sidecar (`initContainers` +
  `restartPolicy: Always`, hence k8s 1.29+): the engine probes the kvd socket once
  with a 5 s timeout and refuses to start if nothing answers, so "kvd first" has to
  be a scheduling guarantee rather than a convention.
- **The recipe's `--io-mode direct` rationale, corrected.** The kvd storage
  classifier now falls back to a `major:minor` walk through sysfs when `lsblk` cannot
  name a device, which is the exact failure the recipe cited as its reason for
  pinning the flag. The flag stays — it is the validated value and a
  misclassification is a silent 4× — but it is a pin now, not a workaround.
- **`examples/glm5.2_gfx942/` — new docker + shell recipe.** `env.sh` plus five
  launch scripts (etcd, optional kvd daemon, both SGLang legs, the Infera router),
  then `verify.sh` and `bench.sh`. Every failure this stack produces silently gets an
  assertion: an arithmetic answer only reachable through an intact prefix, a non-zero
  block count in the router's own pick log, `spec_accept_length` on the decode leg.
  kvd ships here but defaults to `KVD=0`, which is the measured-best shape for this
  workload rather than a hedge; `env.sh` refuses a `KVD` value that is neither 0 nor
  1, since `KVD=true` would otherwise skip the daemon and produce a baseline run
  wearing the kvd name.
- **The manifest's chunk and MTP shape, corrected.** It landed with the docker
  recipe's pre-sweep values while claiming they were identical, and both are the
  sweep's largest levers: `--chunked-prefill-size` 131072 → 8192 (16,384 → 1,024 per
  rank, since DP-attention splits the value `dp_size` ways) and EAGLE 3/1/4 → 5/1/6,
  on both legs, because SGLang rejects a disaggregated pair whose speculative configs
  differ. Section 5's kvd A/B also stops saying the measurement is pending: `KVD=1`
  ran 12.0% slower with `gets_total = 0` — 100.8 GB written to L3 and not one page
  read back — because the 54 GB/rank device pool had already taken ~100% of the reuse
  that trace offered.

Out of scope, deliberately: the agentic multi-turn benchmark that produced the
figures above. `bench.sh` here is `sglang.bench_serving` on random prompts, which
sizes the deployment and reproduces none of them; the trace harness is being
generalised into a standalone tool for a later PR.

## Validation

Run against the live 2 × MI300X deployment:

- `verify.sh` — 5/5 checks pass (both legs registered, correct answer through the PD
  pair, 1561 blocks hashed on the largest router pick, `spec_accept_length` present
  on the decode leg, Mooncake RDMA transport in the decode log).
- `bench.sh` — completes and writes `results/<tag>.{json,log}`; the ~0% cache-hit
  line is expected on a random dataset and is documented as such.
- kvd paths — a real daemon brought up by `launch/launch_kvd.sh` on a scratch socket
  and directory with small quotas: readiness probe and `io_mode` readout both behave.
  `KVD=0` is a clean no-op, `KVD=true` is rejected with exit 1, and `KVD=1` without a
  daemon makes `launch_prefill.sh` exit before it touches the running engines.
- All 12 scripts pass `bash -n`; trailing whitespace, end-of-file newlines and file
  sizes are clean against the repo's pre-commit hooks.

# Checklist:

- [ ] I have added tests that prove my fix is effective or that my feature works — N/A, no library code changes; validation is above
- [ ] New and existing unit tests pass locally with my changes — N/A for the same reason


kzjeef (approved):
please add one k8s config receipt too.

jiejingzhangamd (approved):
please add the k8s config too.

PR #98, opened 2026-08-07
fix(router): index only the attention KV-cache groups in kv-aware routing

jiejingzhangamd:
## The failure

kv-aware routing against **Kimi-K3 on two mixed workers** logged `cache_hits=0`
on every decision and pinned all traffic to one worker. The second node's GPUs
sat idle. Not a degraded hit rate — half the fleet.

## Cause

vLLM emits one `BlockStored` **per KV-cache group**, and only the attention
group pairs one hash with one block. Kimi-K3 is hybrid (24 MLA + 69 KDA); the
KDA groups run prefix caching in `align` mode, which nulls all but one block
per step, so those events carry `token_ids` spanning the whole chunk and a
single surviving hash — with nothing on the wire saying which block it covers.
Captured off a live worker (`block_size 768`, `--max-num-batched-tokens 4096`,
so chunked prefill splits at 3840):

| `kv_cache_spec_kind` | `group_idx` | `token_ids` | `block_hashes` |
|---|---|---|---|
| mamba | 0 | 3840 | 1 |
| mamba | 1 | 3840 | 1 |
| mamba | 2 | 3840 | 1 |
| mla_attention | 3 | 3840 | 5 |

Indexing all four is worse than indexing none. vLLM mixes **no group id into
the block hash**, so at equal block sizes a Mamba hash collides with an
attention hash and overwrites its entry in `map[engine_hash] -> router_hash`.
The next chunk then resolves its parent to the wrong node and everything after
it hashes off a poisoned chain. Which group won depended on intra-batch
emission order, which vLLM does not promise.

Upstream: [vllm#44451](https://github.com/vllm-project/vllm/issues/44451) is
open and untriaged, [vllm#44488](https://github.com/vllm-project/vllm/pull/44488)
(block offsets) has no reviews. NVIDIA Dynamo survives hybrid models because it
filters with `is_main_attention()` — which is what this PR does.

## The fix

Filter by `kv_cache_spec_kind`. The two fields the filter needs (`group_idx`,
`kv_cache_spec_kind`) were on the wire all along — our msgspec struct just did
not declare them, so they were discarded before anything could look. Same
change in the Rust decoder.

Two decisions worth reviewing specifically:

- **`mla_attention` is mandatory in the whitelist.** Kimi-K3's attention layers
  are MLA, so a filter written as `== "full_attention"` discards 100% of that
  model's usable events — the same empty view, reached from the other side.
- **The filter fails OPEN when the field is absent** (SGLang, older vLLM). Those
  engines were indexed before this change; a closed default would switch
  kv-aware routing off for them silently.

**A length disagreement no longer drops the event whole.** An earlier revision
of this branch did, and it measurably costs hits: on Qwen3.5-0.8B (same event
shapes) 18/32 requests hit a full prefix with the whole-event drop, versus
22/32 with no filter at all. The view entries *are* correct — chained from a
resolved parent over a contiguous token span. What is not correct is the
hash-to-block pairing, so **the view is indexed and the map is not**. A later
event then misses its parent and is dropped, which under-reports hits instead
of mis-reporting them.

## Tests

`tests/unit/router` **239 passed**; Rust `cargo test` **80 passed**, `cargo fmt --check` clean.

| File | What it covers |
|---|---|
| `test_kv_event_group_filter.py` | the filter per field, asserted through the query path |
| `test_kv_event_kimi_k3_shapes.py` | replays the captured shapes above, two chunks |
| `test_kv_event_e2e_hybrid.py` | **real ZMQ + msgpack dicts** (not our structs), parametrized over both emission orders; asserts matched block count, not view size |
| `test_kv_aware_routing_extremes.py` | routing outcomes decidable by hand — all-to-the-holder, an exact 10/10 split by content, subset-loses-to-superset, cold-fleet spread, load-vs-locality |

Two of these exist because earlier drafts were vacuous. The e2e test builds
events as **plain dicts** because a test that constructs our own structs starts
from the fixed state and cannot catch a field the schema never declared; and it
runs **both** emission orders because with the attention event last, its correct
hashes overwrite the Mamba groups' and the bug disappears by luck.

The extremes file exists because nothing in the suite ever gave the policy a
*choice* — which is precisely why "32/32 to one worker" was green. Verified
non-vacuous against two degenerate policies: replacing the pick with
`targets[0]` fails 8/10, dropping the cache term so only load counts fails 5/10.

## Not claimed

A local A/B on Qwen3.5-0.8B gives **19/32** full hits with this fix versus
**22/32** on stock v0.2.3. I have not explained that 3-hit gap and am not going
to invent a reason for it. Note the local rig is a *dense-path* comparison —
the production zero-hit case is not reproduced by it, and its most likely cause
was a concurrency flaw in the affinity test I was measuring with (rounds fired
concurrently, so repeats raced KV-event propagation). The hybrid-group bug here
is real and independently demonstrated by the replay tests, but it is not
established that it alone explains the production symptom.

### DIFF
--- a/infera/router/kv_event/client.py
+++ b/infera/router/kv_event/client.py
@@ -91,6 +91,8 @@ class WorkerSubscription:
     # Latched: the block-size mismatch is per-subscription, not per-event, so
     # logging it on every event would bury the fact under thousands of copies.
     block_size_mismatch_logged: bool = False
+    length_mismatch_logged: bool = False
+    non_indexable_group_logged: bool = False
 
     def view_for(self, rank: int | None) -> set[int]:
         return self.views.setdefault(rank or 0, set())
@@ -242,7 +244,40 @@ class KvEventClient:
             sub.view_for(rank).clear()
             sub.map_for(rank).clear()
 
+    # vLLM emits one BlockStored per KV-CACHE GROUP. Only the attention groups
+    # pair one hash with one block; see `events.BlockStored`. Mirrors Dynamo's
+    # `is_main_attention()` filter, which is why Dynamo survives hybrid models.
+    #
+    # `mla_attention` is NOT optional here. Kimi-K3's attention layers are MLA,
+    # so `get_kv_cache_spec_kind` reports MLA_ATTENTION and a filter written as
+    # `== "full_attention"` would discard 100% of that model's usable events --
+    # the same empty view this filter exists to prevent, arrived at from the
+    # other direction.
+    _INDEXABLE_SPEC_KINDS = frozenset({"full_attention", "mla_attention", "sink_full_attention"})
+
+    def _is_indexable_group(self, sub: WorkerSubscription, ev: object) -> bool:
+        kind = getattr(ev, "kv_cache_spec_kind", None)
+        if kind is None:
+            # SGLang, and vLLM builds before the field existed. Fail OPEN: those
+            # engines were being indexed before this filter and a closed default
+            # would silently switch kv-aware routing off for them.
+            return True
+        if kind in self._INDEXABLE_SPEC_KINDS:
+            return True
+        if not sub.non_indexable_group_logged:
+            sub.non_indexable_group_logged = True
+            logger.info(
+                "kv events from %s include non-attention groups (first seen: %s, "
+                "group_idx=%s); indexing only the attention group(s)",
+                sub.worker_id,
+                kind,
+                getattr(ev, "group_idx", None),
+            )
+        return False
+
     def _on_block_stored(self, sub: WorkerSubscription, ev: object, rank: int | None) -> None:
+        if not self._is_indexable_group(sub, ev):
+            return
         view, m = sub.view_for(rank), sub.map_for(rank)
         if ev.parent_block_hash is None:
             parent = ROUTER_SEED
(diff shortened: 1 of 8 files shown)


jiejingzhangamd:
Force-pushed: `cargo clippy -D warnings` caught a real defect in my own test code that `cargo test` hid.

A stray `#[test]` ended up above the doc comment of the test I inserted, so `non_attention_groups_are_not_indexed` carried the attribute twice and `stored_then_prefix_hits_match_query_chain` carried none. The listed test count was 66 either way — the double registration made up for the one that had silently stopped running — so nothing looked wrong locally. `cargo test --lib -- --list | grep -c stored_then_prefix_hits_match_query_chain` returns 0 before the fix and 1 after.

Now green locally: `cargo clippy --all-targets -- -D warnings` clean, `cargo fmt --check` clean, 80 rust tests, 239 python.

jiejingzhangamd:
The red e2e jobs are a bad CI host, not this branch. `crs-m2m-cpu-spur-012` drops off mid-run, taking both of its runner instances with it:

| attempt | job | runner | ended |
|---|---|---|---|
| 1 | e2e-mixed (sglang) | **012**-x2 | 03:10:42 |
| 1 | e2e-mixed (vllm) | **012**-x3 | 03:10:42 |
| 1 | e2e-disag (atom) | 013-x0 | 03:20:36 |
| 2 | e2e-mixed (sglang) | **012**-x2 | 03:44:49 |
| 2 | e2e-disag (atom) | **012**-x3 | 03:44:59 |

GitHub annotates every one of them with *"The self-hosted runner lost communication with the server"*, and the step logs 404 (`BlobNotFound`) because the runner never uploaded them. Two independent jobs ending in the same second, twice, on the same physical host. On the re-run `e2e-mixed (vllm)` landed elsewhere and passed — the only one that changed hosts.

The one exception is attempt 1 `e2e-disag (atom)` on 013, which was `1 passed, 1 error` where the error is `srun ... docker logs infera-e2e-etcd-...` timing out after 300s — a hung Slurm command while collecting logs, after the test itself.

Nothing failed an assertion. Every tier that actually issued a request passed, all of them through the router this PR modifies: `e2e-mixed (atom)` in full, `e2e-disag (vllm)` and `(sglang)`, `e2e-mixed (vllm)` on re-run, plus the first model in each of the killed jobs (`sglang gpt-oss-120b`, `vllm Kimi-K2.6-MXFP4`) which logged `[e2e correctness] PASS` before the host died.

@xiaobochen-amd — the worker-side logs are under `/home/xiaobche/infera-cicd-shared-logs/31140987208-{sglang,atom-disag}` if you want to confirm from that side. I have not re-run a third time: each attempt holds a reservation on the shared GPU cluster, and re-rolling for a healthy host does not seem like the right way to spend it. Say the word if you would rather I did.

jiejingzhangamd:
ci failed because no resource. @xiaobochen-amd

PR #116, opened 2026-08-12
fix(rust): defects introduced by #113's own review fixes

weilei0120:
## What this is

Follow-up to [#113](https://github.com/AMD-AGI/Infera/pull/113), which merged
before these landed. Reviewing that PR's fixes found five problems **in the
fixes themselves** — one of them the same class of bug the round was meant to
eliminate.

## The findings

**A sixth byte-slice was left behind.** The search that found the others
matched two literal widths and missed `min(300)` in `open_decode`, which
truncates the same kind of text — a decode worker's error body. Reproduced as a
panic on the request task, which drops the connection instead of returning the
502 it was building.

**Skipping a message restarted the idle window.** The loop added to skip
JetStream acks recomputed the idle timeout per read, so a peer publishing them
could hold a request open indefinitely, and with the total cap off by default
nothing else bounded it. Worse, the comment claimed the opposite — that a
skipped message could not extend a deadline — which is the kind of statement
that stops the next reader from checking. The window is now measured from the
last frame delivered to the caller.

**An empty `ca.crt` became a silent no-CA client.** `from_pem_bundle` skips
unrecognised sections, so an empty or corrupt file returns `Ok(vec![])` where
the single-certificate call had failed outright. Startup then succeeded and
every API call failed on an unknown issuer, far from the cause.

**The kv reconnect backoff only ever grew.** A few flaps during broker startup
pinned it at 30 s for the life of the process, so every later reconnect blinded
kv-aware for that long — the degradation the reconnect loop exists to prevent.

**Tying the bucket watch to the event stream** made a transient bucket error
throw away a healthy subscription and force a full replay. The two are not
equally important: the stream is authoritative and ordered, the bucket only
shortcuts a cold start. Caught by an integration test that failed only under
parallel execution.

## Tests that were passing without running

The broker-dependent tests returned early when `INFERA_TEST_NATS` was unset,
and a test that returns is reported as **passing**. They are the only cover the
timeout-cancel and fail-open fixes have, so in CI they were green assertions
about nothing. They are `#[ignore]`d now — `cargo test` says `4 ignored`, which
is a truthful "not run".

One of them also depended on a stream an earlier run had left on the broker: on
a fresh one it failed every time. It creates its own precondition now.

## Smaller items

- A negative PubAck is recognised as an ack too — it carries an `error` object
  and no sequence, so it was still being handed to the client as the start of a
  body.
- The 429 body is built through serde: a worker id comes from a Pod annotation,
  and a quote in it produced malformed JSON.
- `Worker` derives `PartialEq` instead of a hand-written twelve-field
  comparison. Forgetting a field there is exactly how the prefill bootstrap
  address went stale in the first place.

## Test plan

- [x] `cargo test` — 154 tests, all green on a **fresh** broker (no inherited state)
- [x] `cargo clippy --all-targets` and `cargo fmt --check` clean
- [x] `pytest tests/unit` — 1338 passing
- [x] Each fix verified to fail when reverted
- [ ] CI

Made with [Cursor](https://cursor.com)

### DIFF
--- a/rust/router/src/disagg.rs
+++ b/rust/router/src/disagg.rs
@@ -541,7 +541,7 @@ async fn open_decode(
                         "decode {} error {}: {}",
                         d.worker.worker_id,
                         st.as_u16(),
-                        &txt[..txt.len().min(300)]
+                        truncate_chars(&txt, 300)
                     ));
                 }
                 state.breaker.record_success(&d.worker.worker_id);
--- a/rust/router/src/discovery_k8s.rs
+++ b/rust/router/src/discovery_k8s.rs
@@ -348,7 +348,7 @@ fn handle_pod(pod: &Value, deleted: bool, tracked: &mut HashMap<String, Tracked>
     let existed = tracked.contains_key(&name);
     let unchanged = tracked
         .get(&name)
-        .is_some_and(|t| !t.announced && same_worker(&t.worker, &worker));
+        .is_some_and(|t| !t.announced && *t.worker == worker);
     if unchanged {
         // Pods generate MODIFIED events for things the router does not care
         // about -- status conditions, resourceVersion bumps on every heartbeat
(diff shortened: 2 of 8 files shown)


weilei0120:
Superseded: both commits now live on feat/rust-router-k8s-nats, which is where this work belongs. Reopening from there.

PR #124, opened 2026-08-19
feat(agent_sys): task_graph — task-management substrate for the agent loop

dorado269:
Adds **`task_graph`**, the component that decides *which task runs when* for Infera's agent-driven performance-optimization loop, alongside the existing `env_mgr` under `agent_sys/`.

An AI agent is treated as a function that is not very procedural. A handoff is that function's input or output. The scheduler decides **when** a task runs and nothing else — it never inspects what a task does and never writes handoff state.

Delivered in three reviewed stages: **specification → design → tests & code.**

| | |
|---|---|
| [`docs/spec.md`](../blob/dev.yihou.aiopt.task.graph/agent_sys/task_graph/docs/spec.md) rev. 7 | What the system must do. 35 acceptance criteria |
| [`docs/design.md`](../blob/dev.yihou.aiopt.task.graph/agent_sys/task_graph/docs/design.md) rev. 10 | Files, classes, interfaces, test plan. §13 records every place implementing the spec literally did not work; §14 the questions left open on purpose |

## Design decisions worth naming

- **Eligibility is a query, not a counter.** `check_if_latest_valid(hid)` is re-asked at each decision point, so nothing has to be invalidated when a producer re-runs.
- **Handoffs are versioned append-only lists.** The list index *is* the version, so contiguity is structural and no validator enforces it.
- **Acquisition is atomic and all-or-nothing.** The full declared set is verified before any pool is touched, which makes hold-and-wait deadlock structurally impossible.
- **Renewable and consumable pools differ in three behaviours** — release, persistence, recovery — which is why the split is at the abstract-base level rather than a boolean flag. A reservation is a lease that dies with its process; a settlement is spend and must not be un-spent.
- **One writer per invariant.** `Scheduler._move` is the only thing that assigns `task.status` or mutates a pool.
- **Composition over inheritance, resolved at use time.** No manager imports another manager; `bootstrap.build_registry()` is the sole composition root.

## Layout

`agent_sys/` is a container for two independent components; neither imports the other.

```
agent_sys/
├── pyproject.toml       declares both packages
├── env_mgr/             (unchanged, from main)
├── task_graph/
│   ├── docs/            spec.md, design.md
│   └── *.py
└── tests/{env_mgr,task_graph}/
```

## Testing

```bash
pip install -e agent_sys
pytest agent_sys
```

**423 tests pass** (358 `task_graph` + 65 `env_mgr`), covering all 35 acceptance criteria. Design §11 maps every criterion to a named test.

[...]

PR #132, opened 2026-08-25
feat(agent_sys): the agent work system — specs, designs, and a running implementation

dorado269:
Builds on the `task_graph` substrate already merged in #124. Adds the nine
remaining packages, their specs and designs, and a CLI that runs a task package
end to end.

**Rebased onto `main` and squashed to a single commit.** The base was moved from
`dev.yihou.aiopt.task.graph` to `main`; the 615 development commits collapse into
one. One of them (`aec06fc`, "all spec first version") carried no
`Signed-off-by` and would have blocked this PR on DCO — squashing dissolves it.

## The claim

**A task is a function**, signature `<handoffs, agent>`. Quality is guaranteed by
standardising what crosses the boundary — the inputs and the outputs — rather
than by trusting the executor to be good.

Everything below follows from taking that literally.

## What is here

| package | what it owns |
|---|---|
| [`spec_loader`](agent_sys/spec_loader/) | YAML task packages: schema, asset auto-discovery, `--var` expansion |
| [`handoff`](agent_sys/handoff/docs/spec.md) | the unit of transfer, its content types and item schemas |
| [`validator`](agent_sys/validator/docs/spec.md) | what makes a handoff checkable, and how far a check can be trusted |
| [`agent`](agent_sys/agent/docs/spec.md) | the executor and the backend abstraction |
| [`closure`](agent_sys/closure/docs/spec.md) | the binding of the four |
| [`env_mgr`](agent_sys/env_mgr/docs/spec.md) | everything the system does to the OS: zones, grants, confinement |
| [`monitor`](agent_sys/monitor/docs/spec.md) | observation without authority over task state |
| [`cli`](agent_sys/cli/docs/spec.md) | `agent-sys run --package <dir>` |
| [`task_graph`](agent_sys/task_graph/docs/spec.md) | subgraphs, validation phases, and the fixes the first real runs exposed |

Nine specs, nine designs, and
[`docs/interfaces.md`](agent_sys/docs/interfaces.md) — normative for everything
that crosses a module boundary, and the document that made the seams reviewable
before the code existed.

Two reference packages, `examples/demo` and `examples/demo2`, plus
`examples/demo-broken` as a deliberate negative. A workflow's specs live in a
task package *outside* this repository; these are the bounded exceptions, and a
test enforces that nothing else creeps in.

## Verification

`pytest agent_sys` — **2051 passed, 3 skipped, 4 xfailed**, on this exact tree
after the rebase. 154 test files.

Both reference packages run to completion through the CLI.

## What review should be sceptical about

Written plainly, because it is what a reviewer would otherwise have to find.

- **A passing validator is not a working system.** An earlier end-to-end run
  reported 14/14 tasks, ten validators PASS and exit code 0 over an artefact in
  which every produced solution scored zero. Every check added since carries a
  non-vacuity control — a case that must still fail — because an instrument only
  ever pointed at the good case proves nothing.
- **The permission system defaults to off** in the last commit of this branch's
  history, by the repository owner's instruction, and the CLI prints a
  `PERMISSIONS_DISABLED` banner on every run saying so. A green run here is
  explicitly *not* evidence that the sandbox works.
- **`env_mgr`'s remote path is designed and partly built, with zero production
  callers.** `Ssh` and `DockerExec` exist and nothing constructs them. It is a
  build, not a debug, and it is not claimed as working.

## Not in this PR

Remote execution, and the one-real-task package that exercises it. Those are the
next stage and are tracked in
[`docs/implementation-stage.md`](agent_sys/docs/implementation-stage.md).


agent_sys/agent/docs/spec.md
```diff
@@ -0,0 +1,395 @@
+# Agent — Specification
+
+| | |
+|---|---|
+| Status | Draft, pending review |
+| Revision | 1 — 2026-08-24 |
+| Date | 2026-08-24 |
+| Scope | What an executor declares, what its knowledge must contain, and the backend abstraction |
+| Source | The task definition §6, §7.8; the Infera × Hyperloom kickoff report §2, §5F |
```
dorado269: don't refer to hyperloom, just use infera agent_sys

agent_sys/agent/docs/spec.md
```diff
@@ -0,0 +1,395 @@
...
+
+The SDK's session id and this system's `AgentId` are different things, and the
+backend records the correspondence. `task_graph` spec §3.3 already establishes
+that one agent means one run; the SDK's `session_id`, `fork_session`, and
+`resume` are how a backend re-establishes a run's context after a restart —
+which is the agent's own business (§1.2), not the system's.
+
+---
+
+## 6. Logging
```
dorado269: logging belongs to another system "o11y". and logging is a independant package: it provide backend specific tools to save logs to json file with predefined schema. then some python post-process visitor will read it to, means the gen write process and the post process is decoupled, althrough some real-time post-process is allowed. also the log storage is managed by fs mgr and abey the permission rules system

dorado269: but you  can append this part as a TODO to the tail of the plan. and we will do the spec, research, design, implement process for it later

agent_sys/agent/docs/spec.md
```diff
@@ -0,0 +1,395 @@
...
+record — its id, its binding to a task, the `HandoffRef`s it touched, its
+persistence — is already specified in `task_graph` spec §3.3 and is **not**
+restated here.
+
+```
+task_graph spec §3.3       this document
+────────────────────       ────────────────────────────────
+Agent      the instance    AgentSpec     the kind
+  .id                        name
+  .task_id                   permissions
```
dorado269: permission is a runtime attr
and by default, agent is only allow to access its own in/out handoff, workspace, playground, log place.
actually, as task is strongly bind to a latest agent(main), and a task should have access permissions to all its subtask in its subgraph(if any) so let's say permissions is a versioned task attr, not agent.
and for a very simple first version the path in system is recursive and nested, so that the storage of sub-task is naturally in the parent's folder, that way, hook is easier to implement(you can check and analysis this logic)

agent_sys/agent/docs/spec.md
```diff
@@ -0,0 +1,395 @@
...
+| Date | 2026-08-24 |
+| Scope | What an executor declares, what its knowledge must contain, and the backend abstraction |
+| Source | The task definition §6, §7.8; the Infera × Hyperloom kickoff report §2, §5F |
+| Depends on | [`../../task_graph/docs/spec.md`](../../task_graph/docs/spec.md) §3.3 — the `Agent` record |
+
+---
+
+## 1. Purpose
+
+An agent is **one executor of one task run**. The system's own definition is
```
dorado269: Just mention and clearify: agent spec is the very thing that wrap task spec. task spec is for input/output、goal template
 it only abstract some common interface and property for backend/harness/sdk: for example : 
1. how to interact with agent, start, stop, interrupt, instruct, query.
2. some config、rule、hook、skill.

dorado269: simply put: it wraps every common interface that major/SOTA agent harness provide: claude code/cursor/codex. but let's start at the basic feature and let the extendibility to framework to ensure it, so that we can leave some implementation later and focus on phase 1 of this project.

dorado269: sorry, just rethink, maybe the "simply put" part is for the second level interface "ai agent", the first is just provide some unifined interface towards task runner of the aiopt system inside

agent_sys/agent/docs/spec.md
```diff
@@ -0,0 +1,395 @@
...
+                                │
+                                ▼
+                          AgentBackend  the harness abstraction
+                             history · interrupt · queue · hooks
+```
+
+### 1.1 In scope
+
+- The YAML agent spec: permissions, environment, knowledge, backend.
+- The four mandatory parts of an agent's knowledge.
```
dorado269: we can first give knowledge four types(extendable):
1. few shot
2. runnable
3. official reference
4. expert experiment/suggestion
5. suggested/verify resource source
6. runtime gened: like senera analysis result, compile_command.json

agent_sys/agent/docs/spec.md
```diff
@@ -0,0 +1,395 @@
+# Agent — Specification
```
dorado269, line 1: an agent spec is bound to a task spec
and an ai agent has a beckend implementation list/dict like [key: claude_code_sdk, backend_entry: ...]

agent_sys/agent/docs/spec.md
```diff
@@ -0,0 +1,395 @@
...
+
+---
+
+## 2. Design principles
+
+| # | Principle | Consequence |
+|---|---|---|
+| 1 | **The executor is interchangeable** | AI, human, or program. The interface is the handoff, not the executor |
+| 2 | **The agent node is coarse** | Whatever a backend organises internally is invisible and unmanaged. §4.1 |
+| 3 | **No zero-shot** | Knowledge is mandatory and has four required parts. §3 |
```
dorado269: not hardcode mandatory, but strongly suggestion with warning and a mandatory option for cli running config.
maybe there is a config dispatch and dissemination system, let make the first version very simple: a simple global yaml config with well classified partition, and everyone read it.

agent_sys/agent/docs/spec.md
```diff
@@ -0,0 +1,395 @@
...
+
+---
+
+## 4. The backend abstraction
+
+### 4.1 The system's agent node is coarse
+
+**Whatever multi-agent structure a backend organises internally is invisible to
+this system, and explicitly not managed by it.** The system always considers the
+work handed to a backend to have been done by one agent.
```
dorado269: yes, from this part of system, and we gonna develop another module responsible for check and ensure at it's best effect every worker agent finish its job:
the monitor agent system, that is:
1. every task, has it's monitor thread/runner to maintain the agent's status and  a. answer the outside where it is, how far it goes, if it is alive, if it still in the right place. b. check
2. normally, the task has it's own monitor. but also, the task append its monitor job to a global monitor who is round-robin-ly manage every task is allowed but when handling the monitor job for a task. it only has permission scope of that task (TODO: history of global monitor will leak context.)
3. a monitor have several ways to handle a agent fail to submit it's handoff or stop/pause in the middle way: a. simplest way is a status checker with a simple phase "continue, do it until finish". b. better way is having a agent give simple analysis and dispatch among several decisions like: [push, help(assign a agent to help it ), add input/knowledge, create coordinator, create teamate/rival. change agent spec, escalate to upper scope, report to user]
at first mvp version, we implement simple pusher

you can add this to TODO at the tail of plan

agent_sys/agent/docs/spec.md
```diff
@@ -0,0 +1,395 @@
...
+
+- **Backends are swappable** because none of them can leak structure into the
+  system's model.
+- **The system's observability claims stay honest.** It reports that an agent ran
+  and what it touched, never what it thought — because for a backend with
+  internal subagents, "what it thought" is not a single thing.
+- **The audit record stays finite.** One `Execution` per run, whatever happened
+  inside.
+
+The cost: a backend that spawns twenty subagents looks, from here, exactly like
```
dorado269: no need, it's simply a interface related to o11y. but one thing we can ensure in a very lang term. although a backend agent cam report how many subagent it has, and make every agent history visible like claude code cli, Only interact with main backend agent is allowed[interrupt/talk...]

agent_sys/agent/docs/spec.md
```diff
@@ -0,0 +1,395 @@
...
+A backend is a thin wrapper exposing a uniform surface. The initial
+implementation may be naive as long as the decoupling is real — the interface is
+the asset, not the wrapper.
+
+| Capability | Required because |
+|---|---|
+| `get_history()` | Observability (main spec §2 principle 4), and it is one of the three things an agent needs to be resumable (env_mgr spec §5) |
+| `interrupt()` | Interventionability (§2 principle 5) |
+| `append(message)` | The other half of intervention: steering a running agent without killing it |
+| `set_rule(...)` | Injecting the system's rules and the task's constraints |
```
dorado269: no need for a runtime interface. agent spec should be well prepared with its rules/hook and everything(defined and setup by env_mgr in env part?), the only thing is a helper to convert rules/hooks/skill among different format of different harness. but is another independent module: some thing like agent_harness_back_transform_helper.[let set the default store form as claude code]

agent_sys/agent/docs/spec.md
```diff
@@ -0,0 +1,395 @@
...
+| `add_hook(event, handler)` | The enforcement mechanism for every boundary in the system that is not a scheduler boundary. §5.4 |
+| `start(task, agent)` / `stop(...)` | The `TaskRunner` protocol `task_graph` spec §5 already defines. A backend satisfies it |
```
dorado269: as task uuid is in runtime agent schema. the task parameter is not needed. but start_async should be provide and return a callback will be called when it really start. agent should have some status like pending、deploying、runing、..... something like that. I think the status of task should just be a supperset/wrapper of stack top's agent's status

dorado269: also a wait() candy and a synchronize start()[wrapper of async] candy can be provide.
for more interface. you can research claude code sdk and cursor sdk and tell me what are the most important/crticial interfaces

agent_sys/agent/docs/spec.md
```diff
@@ -0,0 +1,395 @@
...
+`https://code.claude.com/docs/en/agent-sdk/python.md` unless another page is
+named.
+
+| Required capability | SDK surface |
+|---|---|
+| `get_history()` | `get_session_messages(session_id, directory=, limit=, offset=)` → `list[SessionMessage]`; `list_sessions(...)` → `list[SDKSessionInfo]`; `get_session_info(session_id)` |
+| `interrupt()` | `ClaudeSDKClient.interrupt()` — streaming mode only |
+| `append(message)` | Streaming input: `client.query(prompt: AsyncIterable[dict])`, yielding `{"type": "user", "message": {...}}` items; or a further `client.query(str)` on the same session |
+| `add_hook(event, handler)` | `ClaudeAgentOptions(hooks={HookEvent: [HookMatcher(matcher=..., hooks=[cb])]})` |
+| permission gate | `ClaudeAgentOptions(can_use_tool=...)` for the prompt path; a `PreToolUse` hook to gate **every** call. §5.4 |
```
dorado269: about how permission and isolation of agent is implemented on SOTA harness[like claude code/cursor/codex], you should do deepresearch and gen a report. So that we can refer the SOTA design

agent_sys/agent/docs/spec.md
```diff
@@ -0,0 +1,395 @@
...
+
+| Key | Meaning |
+|---|---|
+| `name` | The spec name. Unique in the registry |
+| `description` | What this kind of executor is for |
+| `backend` | Which backend runs it. §4 |
+| `permissions` | §3.2 |
+| `env` | Environment requirements, resolved by `env_mgr` |
+| `knowledge` | §3.3. **Mandatory** |
+| `resources` | Default pool names and amounts a task using this spec declares |
```
dorado269: resources is task level stuff?

agent_sys/agent/docs/spec.md
```diff
@@ -0,0 +1,395 @@
...
+---
+
+## 8. Acceptance criteria
+
+1. An agent spec declaring no few-shot example is rejected at load, and the
+   message names the missing knowledge part.
+2. An agent spec naming an unregistered backend, an unresolvable knowledge
+   handoff, or a nonexistent permission zone is rejected at load; each message
+   names the offending value.
+3. **A backend satisfies `TaskRunner` unchanged.** The `claude-agent-sdk` backend
```
dorado269: backend and task runner is different thing. task runner is a mid-man between task、task-scheduler、agent, pure func/helper. for decouple purpose (you can argue about this design)
backend is a real-thing that run a common define "agent" like claude code agent after the env/workspace/handoff/playground is deployed

agent_sys/agent/docs/spec.md
```diff
@@ -0,0 +1,395 @@
...
+   behaviour, without ending the run.
+7. **A `PreToolUse` hook blocks a write outside the agent's declared zones**, and
+   the agent receives a message saying why. Verified with `allowed_tools`
+   deliberately naming the tool, to demonstrate that the hook — not the allow
+   list — is what gates.
+8. **A producing agent cannot read a validator's checking standard.** The same
+   hook denies it; the denial is recorded at `warning`.
+9. A system-level logging requirement cannot be suppressed by an agent's own
+   logging rules.
+10. **Swapping the backend changes no other component.** The demo graph
```
dorado269: so the agent related material is classified into two kind: with backend format(plugins、rule、hooks), no backend format(like knowledge)
1. for with backend format, store it as claude code standard, and implement convert/transform helper

agent_sys/agent/docs/spec.md
```diff
@@ -0,0 +1,395 @@
...
+11. The system's record of a run contains no prompt text and no reasoning —
+    asserted over the persisted records, not by inspection.
+
+---
+
+## 9. Open questions
+
+| Item | Status |
+|---|---|
+| **The human executor** | §2 principle 1 says a human may be an executor and nothing contradicts it, but no backend implements one. What a human backend's `get_history()` and `interrupt()` mean is undecided — the honest answers are probably "the notes they took" and "ask them to stop" |
```
dorado269: just deploy. and the human will manually modify the status and provide handoff. that is: no initiative action like detect or cyclical action like heartbeat is allowed, the monitor is released form "most of its duty", only maintain the status when been called(or callback called), anyway there will be a human gui or something in the future, but is not our topic now, just ensure the design and interface supports that kind of extend.

agent_sys/agent/docs/spec.md
```diff
@@ -0,0 +1,395 @@
...
+| **The observer** | Main spec §10 leaves this open system-wide. The agent-level part of it: an outside view of whether *this* agent has drifted from its goal or is looping. `get_history()` supplies the data; nothing consumes it |
+| **Backend capability negotiation** | §4.2 lists required capabilities as though every backend has all of them. A backend that cannot interrupt is still useful; whether it declares that, and what the system does when a task needs a capability the backend lacks, is unspecified |
+| **Cost attribution** | The SDK reports cost per session (`ResultMessage`, `max_budget_usd`). The system has a consumable token pool (`task_graph` spec §3.4) that expects a settled figure at completion. The mapping between them is a design-stage question, but the two have different notions of what a "run" is when a session is resumed |
```
dorado269: o11y system should also record some metrics per run of a agent spec, like cost, time cost, resume times, failed times...
mark as TODOs to as a subsystem of o11y

agent_sys/agent/docs/spec.md
```diff
@@ -0,0 +1,395 @@
...
+| **Knowledge freshness** | §3.3 requires official references, and handoff spec §11 notes that knowledge handoffs outlive runs. Nothing detects that a referenced document has changed, which is precisely the staleness the requirement exists to prevent |
```
dorado269: simply, knowledges has its versions/timestamp/checksum.
and there will be a knowledges record system being called everytime the knowledge storage is update.
 but currently we don't care, just implement the version simple version.
mark it TO roadmap doc long-termly

agent_sys/agent/docs/spec.md
```diff
@@ -0,0 +1,395 @@
...
+| **Subagent visibility** | §4.1 makes internal structure invisible by design. The SDK exposes it — `SubagentStart`, `SubagentStop`, `forward_subagent_text`, and `parent_tool_use_id` on session messages. Whether the system should stay blind or record a summary is a genuine design choice, not an oversight |
```
dorado269: as we have the whole runtime thing harness provided to us, it's all about o11y with some visitor design pattern, so that is decoupled. 
but currently we don't care,
mark it TO roadmap doc long-termly

agent_sys/closure/docs/spec.md
```diff
@@ -0,0 +1,178 @@
...
+closure = < handoff spec set, task spec, agent spec, validator set >
+```
+
+It is predefined because the system does not support dynamic task specs (main
+spec §6), so the handoff specs cannot be dynamic either. Everything the system
+will ever run is declared before it runs.
+
+### 1.1 It supplies no semantics of its own
+
+**A closure is an abstract table — a wrapper — and nothing more.** This is the
```
dorado269: also can provide some candy wrapper query func candy

agent_sys/closure/docs/spec.md
```diff
@@ -0,0 +1,178 @@
+# Closure — Specification
```
dorado269, line 1: so it's spec level stuff

agent_sys/closure/docs/spec.md
```diff
@@ -0,0 +1,178 @@
...
+9. The six reference-workflow steps are each expressible as one closure, and the
+   set loads without error (main spec criterion 7).
+
+---
+
+## 6. Open questions
+
+| Item | Status |
+|---|---|
+| **Closure versioning** | A closure is the recording. Changing one changes what "the same workflow" means, and nothing records which version of a closure a past run used. The `Execution` record names the agent and the handoff versions (`task_graph` spec §3.2) but not the closure |
```
dorado269: closure is a code management level/spec level thing, with some load checker, read only func candy. does this answer your question?

agent_sys/closure/docs/spec.md
```diff
@@ -0,0 +1,178 @@
...
+| **Graph-level composition** | §4.1 leaves it out on purpose, and it has to live somewhere. The likely home is the system whole task (`task_graph` spec §3.2.1), which is the only thing that sees every closure in a graph |
+| **Parameterised closures** | The reference workflow runs the same six steps against different models and hardware. Whether that is six closures with runtime parameters, or a closure per configuration, is undecided — and the answer interacts with main spec §6's static-graph constraint, since a parameter that changes the graph's *shape* is a dynamic spec by another name |
```
dorado269: it's different closures because its different task

agent_sys/closure/docs/spec.md
```diff
@@ -0,0 +1,178 @@
...
+| **Sharing between closures** | §5 criterion 7 permits sharing. Nothing says what happens when a shared handoff kind's spec changes: every closure using it is affected, and nothing enumerates them. The handoff registry's reverse index (handoff spec §8.2) answers the validator half of this; the closure half has no index |
```
dorado269: closure is a code management level/spec level thing, with some load checker, read only func candy, it's a compose of four part ? does this answer your question?

agent_sys/closure/docs/spec.md
```diff
@@ -0,0 +1,178 @@
...
+   set loads without error (main spec criterion 7).
+
+---
+
+## 6. Open questions
+
+| Item | Status |
+|---|---|
+| **Closure versioning** | A closure is the recording. Changing one changes what "the same workflow" means, and nothing records which version of a closure a past run used. The `Execution` record names the agent and the handoff versions (`task_graph` spec §3.2) but not the closure |
+| **Graph-level composition** | §4.1 leaves it out on purpose, and it has to live somewhere. The likely home is the system whole task (`task_graph` spec §3.2.1), which is the only thing that sees every closure in a graph |
```
dorado269: it's defined in a task.
more clearify: 
1. a task contains a task graph or nothing(just it self, namely leaf task)
2. every leaf task is bind to a agent runtime (maybe several agent spec/backend when choose to run)
3. every task has its monitor, it can be a thread or a thread with a AI(not using agent term because its a special agent with no handoff/task system...) or a global thread roundrobinly manage it.
4. maybe the task_scheduler can be improved: 1. has the ability to say: one task must start after another's status reach some point. 2. provide "togather/peer" way that ensure two agent is start togather. but mark is to the long-term road map

agent_sys/demo/docs/spec.md
```diff
@@ -0,0 +1,166 @@
...
+
+### 1.1 In scope
+
+- What the demo must demonstrate, and why each item is on the list.
+- What it must not require.
+- The CLI surface and the observable output.
+
+### 1.2 Out of scope
+
+- **The reference workflow.** The six-step optimisation loop is what the system
```
dorado269: yes, but 1. the agent system and it's usage is decouple, its the very principle of the agent_sys. 2, actually everything should be decouple or keep easy able to decouple in long-term evolution and maintenance. its one of the best practice of spec towards agentic coding repo.

agent_sys/demo/docs/spec.md
```diff
@@ -0,0 +1,166 @@
...
+| **No hand-editing to run twice** | It is idempotent, or it cleans up after itself |
+
+The one unavoidable external dependency is API access for item 5. The demo must
+**fail with a clear message** when credentials are absent — not a stack trace,
+and not a silent fallback to a fake, which would make the run prove less than it
+appears to.
+
+---
+
+## 4. The CLI surface and output
```
dorado269: so another big job is to implement a cli interface for the whole system, which receive a global task. a config yaml, some cli options, and run the whole thing.
Oh.. a lot things to do really

agent_sys/demo/docs/spec.md
```diff
@@ -0,0 +1,166 @@
...
+12. The machine-readable output is sufficient to assert criteria 2–8 without
+    parsing prose.
+
+---
+
+## 6. Open questions
+
+| Item | Status |
+|---|---|
+| **What the demo's task actually does** | §2.1 is illustrative. Something small, verifiable, and not contrived is wanted — a real check over a real artefact, so the validator has something to be honestly `strong` about |
```
dorado269: research and choose/generate one by your self.

agent_sys/demo/docs/spec.md
```diff
@@ -0,0 +1,166 @@
...
+| **Model cost** | One small `claude-agent-sdk` call per run is the intent, but the SDK's default is Claude Code's own tool set and system prompt, which is not small. Which model, which tools, and which effort level is a design-stage decision with a cost attached |
```
dorado269: not a problem. I will provide the claude api key and endpoint for you in config. the env_mgr should have a submodule to setup up claude code sdk for me,

dorado269: cost is not a concern here, just go straight forward.

agent_sys/demo/docs/spec.md
```diff
@@ -0,0 +1,166 @@
...
+| **Where the demo lives in the package** | A `demo` package with a console script, a `tests/demo/`, or an `examples/` directory. This affects whether the demo is covered by CI, which decides whether it stays working |
```
dorado269: an `examples/` , CI is another thing when the system level coding/implement phase is execing

agent_sys/demo/docs/spec.md
```diff
@@ -0,0 +1,166 @@
...
+| **Whether the demo doubles as an integration test** | It should — it is the only thing that exercises every component together. But criterion 11 requires it to fail without credentials, and a CI job without credentials would then fail. Splitting it into a credentialed and an uncredentialed half is the obvious answer and has not been specified |
```
dorado269: leave it to decide later. record it some where like short-term todo.md

agent_sys/docs/spec.md
```diff
@@ -0,0 +1,390 @@
...
+output verified before the next one starts.**
+
+The research appendix behind this system (kickoff report §5A) is a teardown of an
+existing implementation that got this backwards. Its finding, in one line:
+
+> The absence of cheap gates pushed the entire verification burden onto the most
+> expensive gate, so dead ends and live ones cost the same.
+
+Its most-cited concrete failure is the validator problem this system exists to
+solve: correctness was decided by grepping the agent's own report for the string
```
dorado269: do not detail this

agent_sys/docs/spec.md
```diff
@@ -0,0 +1,390 @@
...
+
+- **What an agent does internally.** If a backend organises its own multi-agent
+  structure, the system does not see it and does not manage it (§4.4).
+- **The runtime environment of the program under test.** sglang, vllm, and infera
+  environments and their before/after consistency belong to the handoffs and
+  validators that name them, not to `env_mgr` (see `../env_mgr/docs/spec.md` §7).
+- **Frameworking the test code a validator runs.** The validator system specifies
+  where a check plugs in, not how the check's own code is organised; that belongs
+  to the owning external test system (see `../validator/docs/spec.md` §6).
+- **Dynamic task graphs.** §6 states the record-and-replay scope and its cost.
```
dorado269, line 74: roadmap

agent_sys/docs/spec.md
```diff
@@ -0,0 +1,390 @@
...
+- The **task graph**: which task runs when, including subgraph nesting.
+- The **agent**: what an executor must declare, and the backend abstraction that
+  keeps the system independent of any one harness.
+- The **closure**: the predefined binding of the four.
+- The **environment**: all interaction with the operating system — storage,
+  workspaces, permission zones, local↔remote mapping.
+
+### 1.3 Out of scope
+
+- **What an agent does internally.** If a backend organises its own multi-agent
```
dorado269, line 65: o11y in roadmap. but still, the system don't care backend generally because it decoupled

agent_sys/docs/spec.md
```diff
@@ -0,0 +1,390 @@
...
+| 2 | **Develop against the validator** | The goal is not "the output looks right" but "the output passes its validator". The validator is specified first and the implementation second |
+| 3 | **`<context, worker>`** | A worker is `<executor, knowledge, rules>`; a context is `<content, protocol, validation programs>`. Every part has a clear, complete, checkable input and output, and does only its own job |
+| 4 | **Observability** | Levelled logging, an outside observer that can see whether an agent has drifted from its goal or entered a loop, and a final working-process summary. A system that can be observed can be repaired |
+| 5 | **Interventionability** | An agent can be interrupted and appended to; a control surface can abort |
+| 6 | **Risk has an exit** | What an agent cannot decide, it reports. Some reports may not be self-issued, which is why §5.2 exists |
+| 7 | **Measurable** | A scoring mechanism exists. Where two results are equal, per-token and per-time efficiency break the tie |
+| 8 | **Composable and pluggable** | The optimisation flow does not change, so each new external tool is integrated through a thin wrapper against a standard interface |
```
dorado269: first principles. highloght. add keyword "decoupled"

agent_sys/docs/spec.md
```diff
@@ -0,0 +1,390 @@
...
+
+Eight principles, adopted from the kickoff report §2 and restated here as
+system-level constraints. They are listed in the order in which they decide
+arguments.
+
+| # | Principle | Consequence |
+|---|---|---|
+| 1 | **Reproducible or it did not happen** | Any deliverable, conclusion, or performance number rests on a reproducible basis. A result that cannot be re-obtained is not a result |
+| 2 | **Develop against the validator** | The goal is not "the output looks right" but "the output passes its validator". The validator is specified first and the implementation second |
+| 3 | **`<context, worker>`** | A worker is `<executor, knowledge, rules>`; a context is `<content, protocol, validation programs>`. Every part has a clear, complete, checkable input and output, and does only its own job |
```
dorado269: maybe some keyword of this should be update as the spec of actual architecture using more acturate terms.

agent_sys/docs/spec.md
```diff
@@ -0,0 +1,390 @@
...
+| 7 | **Measurable** | A scoring mechanism exists. Where two results are equal, per-token and per-time efficiency break the tie |
+| 8 | **Composable and pluggable** | The optimisation flow does not change, so each new external tool is integrated through a thin wrapper against a standard interface |
+
+Three further rules are inherited from `task_graph` and apply system-wide because
+they turned out not to be local:
+
+| # | Principle | Consequence |
+|---|---|---|
+| 9 | **Composition over inheritance** | Inheritance appears only where two things genuinely differ in behaviour. Everything else is a `Protocol` resolved by name |
+| 10 | **One fact, one place** | Where two operations mean the same thing, one is expressed in terms of the other. This is not licence to collapse two genuinely different concerns |
```
dorado269: add some rules like: use exist open-source wheel as possible, do not invent wheel before research or there is one

agent_sys/docs/spec.md
```diff
@@ -0,0 +1,390 @@
...
+changes. If that becomes painful, the fix is a shared *loader* the four
+registries call — not a shared registry.
+
+### 4.2 Spec versus instance, in one line each
+
+| Object | The spec says | The instance is |
+|---|---|---|
+| handoff | what this *kind* of artefact must carry | one slot in one graph, with its versions |
+| validator | how this *kind* of check is performed | one run of that check against concrete handoffs |
+| task | what this *step* is: its handoffs and its agent kind | one node in one graph, with its execution history |
```
dorado269, line 599: a task is allowed to have no ai/human agent but simple executable or code/thread simply.

agent_sys/docs/spec.md
```diff
@@ -0,0 +1,390 @@
...
+| agent | what this *kind* of executor knows and may touch | one executor bound to one task run |
```
dorado269, line 600: add this to TODO:
a human interactive interface make this simple: a human get a deploy workspace and start a claude-cli/cursor, finish the job, and call interface to formalize it output and submit it.(not guess, reading inside agent_sys architecture/design, and do some copy/paste dirty work)

dorado269: so that human can join this loop happily

agent_sys/docs/spec.md
```diff
@@ -0,0 +1,390 @@
...
+---
+
+## 5. Two authority boundaries
+
+Everything else in this document is structure. These two are the load-bearing
+constraints, and both exist because the alternative has been observed to fail.
+
+### 5.1 The scheduler decides *when*, never *what*
+
+The scheduler owns task state and never writes handoff state. It asks one
```
dorado269: I thinke the scheduler owns nothing according to the current implementation? Am I wrong?

agent_sys/docs/spec.md
```diff
@@ -0,0 +1,390 @@
...
+| Agent backend | **claude-agent-sdk** | Satisfies every capability the backend abstraction requires — history, interrupt, message-queue append, hooks, permission callback. See agent spec §5 for the mapping, verified against the SDK reference |
+| Test runner | **pytest** | Already a dev dependency |
+
+| Need | Rejected | Why |
+|---|---|---|
+| Scheduling | Prefect, Hatchet, Temporal, Ray, Airflow, Slurm | Each is a platform whose scheduling core is not separable; adopting a server to obtain one primitive. Recorded in full in `task_graph` spec §9 |
+| Graph algorithms | networkx, `graphlib` | The only graph question asked at dispatch time is whether one task's inputs are valid. `graphlib.TopologicalSorter` additionally refuses nodes after `prepare()`, and this graph grows at runtime |
+| Templating for validator blanks | Jinja2 | Deferred, not rejected: whether a blank is filled by text substitution or by a typed config value is a design-stage question. The spec constrains *what* a blank is, not how it is filled |
```
dorado269: personally I don't like jinja, it's old and complex. using yaml override like k8s system or jsonnet

dorado269: but still its defered

agent_sys/handoff/docs/spec.md
```diff
@@ -0,0 +1,384 @@
...
+- Locality independence — what makes a handoff replayable on another machine.
+- The YAML spec file and its registry.
+
+### 1.2 Out of scope
+
+- **The versioned slot** — `task_graph` spec §3.1.
+- **How a validator works** — [`../../validator/docs/spec.md`](../../validator/docs/spec.md).
+  This document specifies only that a binding exists and what a handoff must
+  declare about it.
+- **Where large payloads physically live.** `task_graph` spec §8.2 leaves this
```
dorado269: just treat v1 handoff storage implementation as the fs system with folder trees. actually it do not have much choice[object store/sql], so the interface is very clear

dorado269: it's a key:value thing, and it's versioned, and the permission management is tree like.

agent_sys/handoff/docs/spec.md
```diff
@@ -0,0 +1,384 @@
...
+```
+
+### 4.1 What each tag changes
+
+| Tag | Storage | Permission | Retention |
+|---|---|---|---|
+| `fixed.required` | Handoff storage | Named in the consuming agent's permission list | Kept for the life of the graph run, and archived with it |
+| `fixed.optional` | Handoff storage | Same | Same |
+| `addons.temp` | Playground | Reachable by the injecting and receiving agent only | Discardable when the run ends |
+| `addons.knowledge` | Handoff storage, in a long-lived domain | Broadly readable; narrowly writable | Outlives every run. Versioned like any other handoff |
```
dorado269: a separate handoff storage: knowledge-handoff storage: reuse implement, but to instance in filesystem/other storage interface

agent_sys/handoff/docs/spec.md
```diff
@@ -0,0 +1,384 @@
...
+| Is each validator `strong` or `weak`, and is that honest? | The taxonomy is only useful if the labels are accurate. See validator spec §5 |
+| Does any validator depend on the producer? | If it does, §5.2 of the main spec is violated and the kind is rejected |
+
+The kickoff appendix supplies the failure this guards against directly: a
+correctness gate whose tolerance table permitted 10% of elements to mismatch and
+still reported PASS. It had a validator. Its coverage was the problem.
+
+### 5.3 The binding is recorded on both sides
+
+A handoff spec names its validators; a leaf validator names the handoff kinds it
```
dorado269: the thing is, a validator may need more than one hand-off, and for hand-off may have several validator. notice that.

agent_sys/handoff/docs/spec.md
```diff
@@ -0,0 +1,384 @@
...
+binds to. The redundancy is deliberate and is specified — including which side
+wins on conflict — in [`../../validator/docs/spec.md`](../../validator/docs/spec.md) §4.
+
+---
+
+## 6. Where the payload lives
```
dorado269: v1: filesystem, treed folder. well structed, with prefix/substr/suffix in folder name for simple permission hook. clear enough?

agent_sys/handoff/docs/spec.md
```diff
@@ -0,0 +1,384 @@
...
+path works, a content-addressed store works and would supply the digest for
+free — and a store that rewrites content does not.
+
+---
+
+## 7. Locality independence
+
+**A handoff contains no local information.** It declares its hardware and
+software dependencies, its container image if it needs one, and nothing about the
+machine that produced it. Any machine satisfying the declaration can consume it.
```
dorado269: no, it should record more detail so it's trackable.
but it should be movable.
let say we don't care movable currently, and add this to roadmap

agent_sys/handoff/docs/spec.md
```diff
@@ -0,0 +1,384 @@
...
+|---|---|
+| **Digest canonicalisation** | §3.2 fixes the algorithm and the scope but not the serialisation. Key order, float formatting, and line endings all move the digest. JSON with sorted keys is the obvious answer; it is a design-stage decision and is called out here so it does not get made by accident |
+| **Payload storage** | Inherited from `task_graph` spec §8.2 and narrowed by §6 only. Still open |
+| **`addons.temp` retention** | "Discardable when the run ends" is not "discarded". Nothing sweeps a playground, and an agent's playground is part of what makes it resumable (env_mgr spec §5), so the sweep cannot be unconditional |
+| **Knowledge handoff provenance** | A knowledge handoff outlives every run, so `producer_task_id` may point at a task that no longer exists in any live graph. Whether that reference must remain resolvable — and what archives it if so — is unspecified |
+| **Cross-handoff schema references** | A validator that checks two handoffs against each other (validator spec §3) needs to name fields in both. Whether `content_schema` may `$ref` another kind's schema is undecided; permitting it couples kinds, forbidding it duplicates definitions |
```
dorado269: yes, use handoff uuid for lookup & jsonpath inside if needed

agent_sys/handoff/docs/spec.md
```diff
@@ -0,0 +1,384 @@
...
+---
+
+## 11. Open questions
+
+| Item | Status |
+|---|---|
+| **Digest canonicalisation** | §3.2 fixes the algorithm and the scope but not the serialisation. Key order, float formatting, and line endings all move the digest. JSON with sorted keys is the obvious answer; it is a design-stage decision and is called out here so it does not get made by accident |
+| **Payload storage** | Inherited from `task_graph` spec §8.2 and narrowed by §6 only. Still open |
+| **`addons.temp` retention** | "Discardable when the run ends" is not "discarded". Nothing sweeps a playground, and an agent's playground is part of what makes it resumable (env_mgr spec §5), so the sweep cannot be unconditional |
+| **Knowledge handoff provenance** | A knowledge handoff outlives every run, so `producer_task_id` may point at a task that no longer exists in any live graph. Whether that reference must remain resolvable — and what archives it if so — is unspecified |
```
dorado269: for knowledge kind, it just use handoff term. do not enter the whole task-agent-hand-off system forcely. so the checker can be loose, validator can be just schema and intra-consistance

agent_sys/handoff/docs/spec.md
```diff
@@ -0,0 +1,384 @@
...
+**What it is and is not for.** It detects accidental corruption and casual
+tampering, and it makes "is this the same artefact I validated?" answerable.
+It is not a security boundary: an agent that can write a handoff can write its
+digest too. The boundary that matters is §5.2 of the main spec — the producer
+cannot grade its own output — and no hash substitutes for it.
+
+### 3.3 Content is executable, result, or both
+
+```
+content
```
dorado269, line 76: contents is a readme.md(must have some section, a guideline/skill/hook should be provided somewhere else in the system: agent env?) + a dict:
key is either defined in the handoff schema or gened by agent runtimely accordingly.
value can be several types:
1. reproducible: readme.md、script/command、result、code(list of repo、self code、 patchwork or concate、 merge.. lives in scripte. here is materials that well structured)、env（hardware/software/repos, skill/hook should be added for agent env）、logs、watchout
3. code: readme.md, codes(use skills to ensure the boundary/interface it clear, and validators to ensure its somewhere testable)
4. structured text: readme.md, text.json/yaml/xml...,(maybe several files) schema: if there is one.
5. text: readme.md, content(maybe several file)

add this to TODO:
for every type content of hand off, there need be a agent skill/rule/hook set to make it.

add this to roadmap:
there should be a standard/rule/checker for review adding parts to agent_sys
every module should have some hard/soft standard to guaranteeing the quility of the code or an obj spec.

maybe there is more, but this is the most I can get in my thoughts now.

agent_sys/validator/docs/spec.md
```diff
@@ -0,0 +1,581 @@
...
+
+A handoff is only a contract if something checks it. **Validators are the sole
+standard by which a handoff is judged**, which is why this is the largest
+document in the set and why the review bar for a validator is higher than for the
+thing it checks.
+
+The system exists because the alternative was measured. The kickoff report's
+teardown of an existing implementation (§5A) found:
+
+> `correctness_passed` came from one of two places: a CLI override, or a regex
```
dorado269: no need to mention shortage of other system to much. just simply explain what's our goal.

dorado269: and why our design make things better for our domain/goal

agent_sys/validator/docs/spec.md
```diff
@@ -0,0 +1,581 @@
...
+> verification burden onto the most expensive gate**, so dead ends and live ones
+> cost the same.
+
+This document specifies the cheap gates.
+
+### 1.1 In scope
+
+- What a validator is, structurally, and why it is a task.
+- The three elements: input, process, result.
+- Templates, blanks, and bounded recursion.
```
dorado269: yes, don't make things complex, just ensure:
1.  avoid copy paste of similar validator
2. give an open for extend and plugins framework

agent_sys/validator/docs/spec.md
```diff
@@ -0,0 +1,581 @@
...
+  must be honest.
+- Folder conventions, tags, and the registry.
+- The two principles that decide the hard cases.
+- A worked example over the reference workflow.
+
+### 1.2 Out of scope
+
+- **Frameworking the test code itself.** §6.
+- **How a handoff is versioned or stored** — handoff spec, `task_graph` spec §3.1.
+- **When a validator task runs** — that is scheduling, and the scheduler decides
```
dorado269: 1. add a very simple schedule policy as default: stack like/depth wise: that is, always unpack the sub task graph of the stack.peak/top() task if can and execute able from top.
this way. a (input validation->main->output validation) task flow will be continuous.

agent_sys/validator/docs/spec.md
```diff
@@ -0,0 +1,581 @@
...
+
+---
+
+## 2. Design principles
+
+| # | Principle | Consequence |
+|---|---|---|
+| 1 | **The producer never grades its own output** | Enforced by hook, not by convention. §8.1 |
+| 2 | **Anything that can be code is code** | The agent fills blanks in a fixed procedure; it does not invent the procedure. §8.2 |
+| 3 | **A validator is a task** | It runs through the same scheduler, gets the same audit record, and is subject to the same authority boundary as any other work. §3 |
```
dorado269: add this to roadmap:
agent env can be resused directly or simply modify(be careful, don't not blur the standard of the whole system design principle)

agent_sys/validator/docs/spec.md
```diff
@@ -0,0 +1,581 @@
...
+| 4 | **The standard is external** | Both the checking logic and the checking criterion come from outside the thing being checked. §5 |
+| 5 | **Label honestly** | The `strong`/`weak` taxonomy is only useful if the labels are accurate. A weak check labelled strong is worse than no check, because it stops anyone looking further. §5.4 |
+| 6 | **Cheap gates before expensive ones** | The reason the system exists. A validator that costs a GPU-hour is a last resort, not a first line |
+
+---
+
+## 3. A validator is a task
+
+**A validator is a special kind of task spec.** Not a callback, not a plugin, not
```
dorado269: actually you know what, I think validator is a special task, but it's invisible from scheduler side. let's just say that a task have three phase: 1. input validations. 2. main/subgraph. 3. output validations.
that is saying: 1. validation are not visible by scheduler 2.. validation still have other feature of task like: a. validation still need a new clear agent env, also can have ai agent inside. (add task agent reuse env feature to TODO.md) b. validation's input is still handoff, but it don't have output handoff, it will call handoff update_validation_status(version_handoff, xxx) to do things(this update is persistable(have a yaml to record & maintain) is a but not excluded from the handoff checksum)
And phase can be skipped by config. also can be skipped if already validated by someone else(make this a cli switch[something like validation strict level]).
also I think every versioned handoff runtime should have a record history list: for every validator, the history executable result(and some other info: for example: validated in what task/versioned agent, validation envs[can be get from the agent]).

agent_sys/validator/docs/spec.md
```diff
@@ -0,0 +1,581 @@
...
+- **It is scheduled**, so it competes for resources like anything else. A
+  validator that needs a GPU declares one and waits for it; a validator embedded
+  as a callback inside another task would take its host's lease implicitly.
+- **It is audited.** `Execution` records which agent ran the check, against which
+  input versions (`task_graph` spec §3.2). "Which version of the checker passed
+  this artefact" is answerable without new machinery.
+- **It runs in its own context.** Which is §8.1, the reason the whole document
+  exists.
+
+The cost is that a check is a graph node, so a graph with many validators has
```
dorado269: new decision[see above]: not a graph node

agent_sys/validator/docs/spec.md
```diff
@@ -0,0 +1,581 @@
...
+|---|---|
+| **Single node.** A validator contains no subtasks | A check that expands into a subgraph is a workflow, and a workflow needs its own validators. The recursion has to stop somewhere, and this is where |
+| **Its own input validation is empty** | Otherwise validating a validator's inputs requires a validator, without bound |
+| **Consumes one or more handoffs; produces `dict[HandoffId, bool]`** | The result is a verdict per handoff, not per validator. One validator checking three handoffs returns three verdicts |
+
+### 3.2 Why a task and not a callback
+
+Three things fall out for free, and each would otherwise have to be built:
+
+- **It is scheduled**, so it competes for resources like anything else. A
```
dorado269: no

agent_sys/validator/docs/spec.md
```diff
@@ -0,0 +1,581 @@
...
+  validator that needs a GPU declares one and waits for it; a validator embedded
+  as a callback inside another task would take its host's lease implicitly.
+- **It is audited.** `Execution` records which agent ran the check, against which
+  input versions (`task_graph` spec §3.2). "Which version of the checker passed
+  this artefact" is answerable without new machinery.
+- **It runs in its own context.** Which is §8.1, the reason the whole document
```
dorado269: and may have it's own agent

dorado269: and it's life status.
this make a 'real' task status more complex: means input validating/output validating is a status. but it worths

agent_sys/validator/docs/spec.md
```diff
@@ -0,0 +1,581 @@
...
+
+### 4.4 The binding, and which side wins
+
+A leaf validator names the handoff kinds it binds to; a handoff kind names its
+validators (handoff spec §8.1). Both sides are recorded, deliberately.
+
+- **Why both:** each direction is needed at a different moment. Loading a handoff
+  kind must resolve its validators; reviewing a validator must show what it
+  covers. Deriving either from the other is a full scan of the other registry.
+- **On conflict, the handoff kind wins,** and the mismatch is **reported, not
```
dorado269: should be on conflict. double side link just for convenience. crash on conflict

agent_sys/validator/docs/spec.md
```diff
@@ -0,0 +1,581 @@
...
+2. There is a result and a ground truth, or a way to produce both.
+3. The risk closes: a failure is detectable within the loop.
+
+**`long-term strong`** is a sub-case: the same rigour, but the verification chain
+only closes at the end of the run, in combination with other handoffs. A trace's
+plausibility judged after the optimisation it justified has been measured is
+long-term strong — quantified, but not answerable at the moment the trace is
+produced.
+
+**`weak`** — no clear criterion, or none available on a short loop:
```
dorado269: add to roadmap: there should be a review guard/a hook/a code part to check the metadata and the content/implementation of the validation's consistency

agent_sys/validator/docs/spec.md
```diff
@@ -0,0 +1,581 @@
...
+1. A validator spec declaring a subtask is rejected at load.
+2. A validator spec declaring a non-empty input validation is rejected at load.
+3. A `.leaf.` validator declaring blanks is rejected at load, and the message
+   says which blank.
+4. Template composition exceeding the configured depth cap is rejected **at
+   load**, with the composition chain in the message — not at run time.
+5. A validator spec with no `strength` tag is rejected. There is no default.
+6. A validator returns `dict[HandoffId, bool]` with one entry per input handoff —
+   verified for a validator taking three handoffs.
+7. **A validator task is dispatched by the scheduler like any other task**: it
```
dorado269: no

dorado269: no in v1, can be discussed later in roadmap

agent_sys/validator/docs/spec.md
```diff
@@ -0,0 +1,581 @@
...
+9. A validator whose logic was written by the agent under test is rejected — the
+   check is structural (whose permission zone the logic lives in), not
+   declarative.
+10. The two-way binding is consistent after loading both registries; a deliberate
+    mismatch is reported, the handoff kind's side wins, and nothing is silently
+    rewritten.
+11. The registry answers "who uses this validator" and "who has used it"
+    separately, and a validator that has never run is distinguishable from one
+    that runs constantly.
+12. An `output_validation` validator runs in the same environment as the task
```
dorado269: No, this is against the very basic and critical rule? 
validation has it's own clear env. the env config can be same as the producer/consumer, but the context is very clear: a flesh agent env and (a flesh agent if needed)
if self is bind to a real agent and the agent env is defined, use it.
otherwise:
the default env config for input validation is using its consumer's.
the default env config of output validation is using its producer's.
otherwise using a predefined global one.

agent_sys/validator/docs/spec.md
```diff
@@ -0,0 +1,581 @@
...
+    validator: numbers computed by code into JSON, rendered by an agent, and the
+    render checked for verbatim copying.
+
+---
+
+## 11. Open questions
+
+| Item | Status |
+|---|---|
+| **Score-typed results** | §3.4. Reserved in the schema, unbuilt. Building it requires first specifying how a threshold is derived from measured run-to-run variance, which is a measurement task, not a specification task |
```
dorado269: not in alpha version, in roadmap maybe

agent_sys/validator/docs/spec.md
```diff
@@ -0,0 +1,581 @@
...
+| **The recursion depth number** | Three, adopted from the task definition's suggestion, is a guess. Nothing has yet needed two levels, so the cap has not been tested against a real library |
```
dorado269: TODO: should have a code review guard

agent_sys/validator/docs/spec.md
```diff
@@ -0,0 +1,581 @@
...
+| **How a cross-handoff input validation names its inputs** | §3.3 puts cross-handoff checks in `input_validation`, and §4.1 says a validator declares its input mapping by handoff kind. When a task has two inputs of the *same* kind, the mapping is ambiguous. Positional? A declared role name? Undecided; related to handoff spec §11's cross-schema reference question |
```
dorado269: alpha version don't care

agent_sys/validator/docs/spec.md
```diff
@@ -0,0 +1,581 @@
...
+| **Dynamic logic provenance** | §5.1's `external_dynamic` — logic produced by an earlier task — is trusted because the producing task is not the checked task. Nothing verifies that at load time, because the graph shape is what makes it true and the registry does not see the graph. A load-time check would need the closure |
+| **Validator versioning** | A validator that changes has re-graded history. Whether a verdict records which validator version produced it, and whether an old verdict is invalidated when its validator changes, is unspecified. `long_term_strong` makes this pressing: the global validator runs against verdicts recorded much earlier |
```
dorado269: not need. sorry for the confusion. long_term_strong means some quality can not be verified/validated once the handoff generated. it must wait for a long loop feedback: like 1. until some downstream task finished. 2.  whole loop/e2e job finished. 3. even further, some external customers feedback

agent_sys/validator/docs/spec.md
```diff
@@ -0,0 +1,581 @@
...
+| **Weak-validator aggregation** | Several weak checks agreeing is more informative than one, and the kickoff report's strongest recommendation on this point is to use a different model family for the reviewer than for the producer, so the two do not share blind spots. Nothing specifies how multiple weak verdicts combine, or whether the system should require model-family diversity |
```
dorado269: I don't think so, it's just a trick can be noted like in the corner of roadmap. it not a design level stuff

agent_sys/validator/docs/spec.md
```diff
@@ -0,0 +1,581 @@
...
+| **Cost-aware ordering** | §2 principle 6 wants cheap gates first, and §7.3 tags cost — but nothing uses the tag. Whether the scheduler should order validators by cost, and how that interacts with it not inspecting content, is open |
```
dorado269: ha? no, nothing should be done about this topic. (why it is here?)

dorado269: from o11y, every task metrics should be standardly record for present and further analysis and opt, but is not hardcode/embeded here and it's not alpha should care.

agent_sys/validator/docs/spec.md
```diff
@@ -0,0 +1,581 @@
...
+> interval.
+
+Specifying a score type before specifying how its threshold is derived would
+reproduce that. See §11.
+
+---
+
+## 4. The three elements
+
+Every validator is `<input, process, result>`.
```
dorado269: change it: every validator is <goal, inputs, implementation, result>

agent_sys/validator/docs/spec.md
```diff
@@ -0,0 +1,581 @@
+# Validator — Specification
```
dorado269, line 1: 想了半天：
2. 我认为validator可以以分为三个维度（考察维度：需要代码化）：
    1. for 完整性/标准度
        1. 有标准的schema等，较好完成，较容易strong化
    2. for 可用度
        1. 可用度，能否给后面的task用、好用不。容易strong化
    3. for 可信度
        1. 往往难以量化或者短期难以量化，难以strong化
        2. 一些可以由e2e系统级别的长反馈给出较为strong的结论。
        3. 另一些可以由内部逻辑一致性自检给出较为strong的结论。
        4. 总之应该是难关。
3. validator spec还应该有一个brief字段一句话说明他想检查那个维度的具体什么
4. 把那个block留白的模板系统给我取了先。写进roadmap. 感觉是过度设计了。
5. 总觉得现在这个validator系统设计的太复杂了，你有什么简化的想法么？或者你上网deepresearch一下validator/evaluator系统都是怎样设计的，怎样解耦的，都是怎样精简的。

agent_sys/env_mgr/docs/spec.md
```diff
@@ -0,0 +1,376 @@
...
+the agent's own working environment, the remote machine where experiments run,
+and the permission boundaries that make the other components' authority rules
+real rather than intended.
+
+The shipped `env_mgr` is a layered environment manager: one YAML recipe drives
+check / dry-run / install / bootstrap across Python, apt, binaries, and Claude
+plugins. That remains, and §9 says what survives unchanged. This document widens
+the remit around it.
+
+**Why one component and not several.** These responsibilities look separable —
```
dorado269: yes but remeber how to remote detection access remote is a knowledge produce by a specical global task, and the remote <-> local filesystem mapping system is also maintained by env_mgr
so general env_mgr can have may modules, some is reusable but generally it's responsible for a env mangement stuff

agent_sys/env_mgr/docs/spec.md
```diff
@@ -0,0 +1,376 @@
...
+- Permission zones and the prefix convention that makes them cheap to check.
+- Remote access.
+- The agent's local and remote environments: workspace, playground, handoff
+  storage.
+- Preparing an agent's environment before it works.
+- Its own metadata and configuration.
+
+### 1.2 Out of scope
+
+**The runtime environment of the program under test.** sglang, vllm, and infera
```
dorado269, line 52: yes, so in alpha version, a versioned handoff whose have some executable content must record it's env at least(mentioned in handoff spec)

agent_sys/env_mgr/docs/spec.md
```diff
@@ -0,0 +1,376 @@
...
+Also out of scope: **an agent's own process and durability** (`task_graph` spec
+§1.2), and **what a backend does internally** (agent spec §4.1).
+
+---
+
+## 2. Design principles
+
+| # | Principle | Consequence |
+|---|---|---|
+| 1 | **The path is the fact** | Permission, storage, and mapping are all expressed against paths, in one component. §1 |
```
dorado269: just for v1 filesystem implemention

agent_sys/env_mgr/docs/spec.md
```diff
@@ -0,0 +1,376 @@
...
+| 2 | **Write narrowly, read broadly** | An agent may not write outside its zone; reads are looser. §4.3 |
```
dorado269: no such thing, explain to me why it's here.

agent_sys/env_mgr/docs/spec.md
```diff
@@ -0,0 +1,376 @@
...
+that agent resumable (§6.1).
+
+### 4.2 The permission zone prefix convention
+
+**A governed path is recognisable from its prefix alone.** Directories are laid
+out flat, with a common prefix that a hook can match without consulting any
+registry:
+
+```
+${root}/agent-handoff-permission-zone-prefix.<zone-id>/...
```
dorado269: zone id use runtime uuid.version directlly

agent_sys/env_mgr/docs/spec.md
```diff
@@ -0,0 +1,376 @@
...
+        └──────── the marker ────────────┘ └── the zone ──┘
+```
+
+The check is two steps, and the split is the point:
+
+1. **Is this a governed region?** Match the prefix. A path that does not carry
```
dorado269: so let's be clear that, the env mgr whole system contains many pieces, and every part itself is consistent, but the mgr is not like a whole process or thread the maintains everything. it may occurs as a provider of  mechanics and tool for others. as long as the every part of  a mechanic logic is consistent.

agent_sys/env_mgr/docs/spec.md
```diff
@@ -0,0 +1,376 @@
...
+
+Flat rather than nested, because a nested layout makes the zone a function of
+depth, and a hook would have to walk the path to find it.
+
+### 4.3 The two directional rules
+
+| Direction | Rule |
+|---|---|
+| **Write** | An agent may not write outside its zones. Local or remote, no exception |
+| **Read** | Looser. A governed path the agent's list does not name is denied; everything else is allowed |
```
dorado269: add risk note of this to roadmap file

agent_sys/env_mgr/docs/spec.md
```diff
@@ -0,0 +1,376 @@
...
+without being asked for is a sync that happens at the wrong moment.
+
+`local` and `remote` may be the same machine. Nothing in the model assumes
+otherwise, and the mapping then has strong strength trivially.
+
+---
+
+## 5. Remote access
+
+Two mechanisms: **ssh** and **docker exec**. Both are specified as ways to reach
```
dorado269: roadmap: kubectl exec

dorado269: roadmap: spur/slurm exec

agent_sys/env_mgr/docs/spec.md
```diff
@@ -0,0 +1,376 @@
...
+1. A domain is registered, reloaded idempotently, and its kind
+   (`handoff_storage` / `playground` / `workspace`) determines its layout.
+2. **A path carrying the zone prefix is recognised as governed from the prefix
+   alone**, without consulting a registry — verified by checking a path whose
+   zone id does not exist.
+3. **A write outside the agent's declared zones is blocked by a `PreToolUse`
+   hook**, and the agent receives a message saying why. Verified with the tool
+   named in `allowed_tools`, to demonstrate the hook and not the allow list is
+   what gates.
+4. **A read into a governed zone the agent's list does not name is denied; a read
```
dorado269: one more clarification: 
1. the prefix should be configed as un-guessable, like a gened hash, so that a. there is no mis-block ungoverned visit. b. the agent can work around it by some trick like concate two substr
2. although hook is not a perfect way, but let implement it try our best: scripted/pythoned/programed workaround will also be blocked

agent_sys/env_mgr/docs/spec.md
```diff
@@ -0,0 +1,376 @@
...
+   outside every governed zone is allowed.** The asymmetry of §4.3 is asserted
+   directly, not inferred.
+5. **A producing agent cannot read a validator's checking standard**, because the
+   standard's zone is absent from its list — the validator boundary
+   (validator spec §8.1) resolved entirely by §4.2.
+6. A weakly-mapped domain does **not** synchronise until sync is explicitly
```
dorado269: make this a standard tool call, actually make the whole remote-local thing a toolcall for the agent part(not a nature language describe) so ai can easily use it

agent_sys/env_mgr/docs/spec.md
```diff
@@ -0,0 +1,376 @@
...
+13. **The shipped recipe and installer machinery is untouched**: `pytest
+    agent_sys/tests/env_mgr` passes unchanged.
+
+---
+
+## 11. Open questions
+
+| Item | Status |
+|---|---|
+| **Playground retention** | §6.1 makes the playground part of what a resume needs, and handoff spec §11 notes that nothing sweeps it. Those pull in opposite directions and the resolution — an explicit end-of-run archive, probably — is unspecified |
```
dorado269: playground is not synced between remote and local, there content can be totally different. playground can be reused after resume becasue is on fs. but the content is not guaranteed(as it's a playground for wip/temp/un-standard/unwashed content)

agent_sys/env_mgr/docs/spec.md
```diff
@@ -0,0 +1,376 @@
...
+| **Zone identity across machines** | §4.5 maps paths between local and remote. Whether a zone id is the same on both sides, or whether the mapping translates it, is undecided. Same-id is simpler and assumes the operator laid both sides out identically |
```
dorado269: same, we even use rsync, why it should be different

agent_sys/env_mgr/docs/spec.md
```diff
@@ -0,0 +1,376 @@
...
+| **Concurrent agents in one domain** | Two agents with the same playground domain will collide. Whether a playground is per-agent by construction, or whether the model needs a lock, is unspecified — and main spec §10's multi-graph question is the same question one level up |
```
dorado269: playground is a peer agent thing for an agent to record its wip/temp/un-standard/unwashed content) like /tmp in linux.

agent_sys/env_mgr/docs/spec.md
```diff
@@ -0,0 +1,376 @@
...
+| **Sync direction and conflict** | §4.5's weak mapping is `rsync`, which has a direction. Which side wins when both have changed is not specified, and "the caller decides" is an answer that will produce a data loss eventually |
+| **Auto-detection scope** | §3.1 prefers auto-detection. What is actually detectable — NFS yes, worktree yes, sync strength probably, remote reachability only by trying — has not been enumerated, and the boundary decides how much configuration an operator must write |
+| **System-level tasks** | §7 concludes that maintaining cluster conventions implies system-level tasks, agents, and handoffs. None is specified. They are the first thing that will need one |
```
dorado269: so give the root permission.

agent_sys/env_mgr/docs/spec.md
```diff
@@ -0,0 +1,376 @@
...
+detect-and-print only. The second of those overlaps §6.1 and is superseded by it;
+the other two stay as recorded.
+
+---
+
+## 10. Acceptance criteria
+
+1. A domain is registered, reloaded idempotently, and its kind
+   (`handoff_storage` / `playground` / `workspace`) determines its layout.
+2. **A path carrying the zone prefix is recognised as governed from the prefix
```
dorado269: 调研分析下
1. 我们这个path机制有没有什么漏洞
2. 有没有什么更好的实现方式。

agent_sys/env_mgr/docs/spec.md
```diff
@@ -0,0 +1,376 @@
...
+---
+
+## 11. Open questions
+
+| Item | Status |
+|---|---|
+| **Playground retention** | §6.1 makes the playground part of what a resume needs, and handoff spec §11 notes that nothing sweeps it. Those pull in opposite directions and the resolution — an explicit end-of-run archive, probably — is unspecified |
+| **Zone identity across machines** | §4.5 maps paths between local and remote. Whether a zone id is the same on both sides, or whether the mapping translates it, is undecided. Same-id is simpler and assumes the operator laid both sides out identically |
+| **Concurrent agents in one domain** | Two agents with the same playground domain will collide. Whether a playground is per-agent by construction, or whether the model needs a lock, is unspecified — and main spec §10's multi-graph question is the same question one level up |
+| **Sync direction and conflict** | §4.5's weak mapping is `rsync`, which has a direction. Which side wins when both have changed is not specified, and "the caller decides" is an answer that will produce a data loss eventually |
```
dorado269: 1. provide peer agent/peer case level sync method, not whole root
2.  sync should be one time job at some point, not any time: task start, create everything in both local and remote, and ensure the local and remote is same.

agent_sys/docs/design.md
```diff
@@ -0,0 +1,884 @@
...
+
+---
+
+## 12. New open questions
+
+Found by this design, and **not** in spec §10.
+
+| # | Question |
+|---|---|
+| **O1** | **Three runtime dependencies are undeclared.** `agent_sys/pyproject.toml` lists `pyyaml`, `packaging`, `pydantic`. Spec §7 adopts **jsonnet** and **jsonschema**, and handoff §5.1 / validator §4.1 both require **jsonpath** lookup — `jsonpath-ng` is neither installed nor declared, and no module has chosen an implementation. Three names must be added and one must first be picked. This blocks step 1 of §10 |
```
dorado269: coding时根据需要加吧。

agent_sys/docs/design.md
```diff
@@ -0,0 +1,884 @@
...
+| **O2** | **`_jsonnet` ships no Linux aarch64 wheel.** Four abi3 wheels — macOS arm64, manylinux x86_64, musllinux x86_64, win_amd64 — so ARM Linux falls back to an sdist build needing a compiler. `rjsonnet` (Rust `jrsonnet`) ships fourteen and gave better error messages in every case tested, but its `import_callback` returns `str` where `_jsonnet` requires `bytes`, so a runtime seam must normalise that. `kapitan` already does this (`select_jsonnet_runtime`), so it is proven. Not built; §3.2's `render` is the one place it would go |
```
dorado269: 你可以假设aarch64甚至不在roadmap里

agent_sys/docs/design.md
```diff
@@ -0,0 +1,884 @@
...
+| **O3** | **Should the import resolver enforce containment?** §3.3. Spec §4.3 permits cross-package reference and says the loader does not interpret layout — the *permissive* branch, and the one with the CVEs. Kustomize resolves symlinks then enforces containment, calls it "an intentional security feature", and names the cost: it **breaks relocatability**. Helm permits escape with a log line — CVE-2025-53547, *"Helm warns of the symlinked file but did not stop execution"*. dbt permits arbitrary local paths and closed the resulting bug `not_planned`. **This is a specification question**, raised here because the design must not pre-empt it |
+| **O4** | **Nothing says spec numbers are IEEE 754 doubles.** `12345678901234567890` renders as `…567168`. Any field needing an exact large integer must be declared as a string, and no schema currently does. Harmless today because no spec field is a large integer; it will not stay harmless |
```
dorado269: 不算问题哈，只是一个具体的package的用法的常见的坑

agent_sys/docs/design.md
```diff
@@ -0,0 +1,884 @@
...
+| **O5** | **`Registry.get` does not list candidates.** §5.4. The new spec registries do, `env_mgr` already does, and `task_graph`'s component registry does not. It is a one-line change to shipped code and belongs to module 4's design, not to this document |
```
dorado269, line 1141: 没太理解，不list candidates那这个get有啥用？

agent_sys/docs/design.md
```diff
@@ -0,0 +1,884 @@
...
+| **O2** | **`_jsonnet` ships no Linux aarch64 wheel.** Four abi3 wheels — macOS arm64, manylinux x86_64, musllinux x86_64, win_amd64 — so ARM Linux falls back to an sdist build needing a compiler. `rjsonnet` (Rust `jrsonnet`) ships fourteen and gave better error messages in every case tested, but its `import_callback` returns `str` where `_jsonnet` requires `bytes`, so a runtime seam must normalise that. `kapitan` already does this (`select_jsonnet_runtime`), so it is proven. Not built; §3.2's `render` is the one place it would go |
+| **O3** | **Should the import resolver enforce containment?** §3.3. Spec §4.3 permits cross-package reference and says the loader does not interpret layout — the *permissive* branch, and the one with the CVEs. Kustomize resolves symlinks then enforces containment, calls it "an intentional security feature", and names the cost: it **breaks relocatability**. Helm permits escape with a log line — CVE-2025-53547, *"Helm warns of the symlinked file but did not stop execution"*. dbt permits arbitrary local paths and closed the resulting bug `not_planned`. **This is a specification question**, raised here because the design must not pre-empt it |
```
dorado269: 不存在路径逃逸的问题哈，1. 就是说一个具体的task package只是一个单独的root level folder，不是一个单独的repo.2.  就算是一个单独的repo, 也肯定有添加授信列表/copy&paste这种方法可以解决，但是这是roadmap的事情

agent_sys/docs/design.md
```diff
@@ -0,0 +1,884 @@
...
+| **O4** | **Nothing says spec numbers are IEEE 754 doubles.** `12345678901234567890` renders as `…567168`. Any field needing an exact large integer must be declared as a string, and no schema currently does. Harmless today because no spec field is a large integer; it will not stay harmless |
+| **O5** | **`Registry.get` does not list candidates.** §5.4. The new spec registries do, `env_mgr` already does, and `task_graph`'s component registry does not. It is a one-line change to shipped code and belongs to module 4's design, not to this document |
+| **O6** | **SPDX headers are inconsistent.** `env_mgr`'s 19 modules carry one; `task_graph`'s carry none. The new packages follow `task_graph`. Somebody should decide it once for the repository rather than per package |
```
dorado269, line 1142: agent_sys之外infera源代码是怎么做的，一样的做法就行。

agent_sys/docs/design.md
```diff
@@ -0,0 +1,884 @@
...
+| **O7** | **Nothing yet sweeps a stale spec.** A registry is loaded once at composition and never invalidated. Reloading a package after editing a spec means rebuilding the registry, which means restarting the system. Acceptable for the alpha — spec §6 makes graphs static — but it interacts badly with a long-running system, and it is the first thing a developer editing a task package will hit |
```
dorado269, line 1143: 谁管这个啊，就算有resume机制，一个大的任务最多也就持续跑一周。spec里一堆旧垃圾又咋了。

agent_sys/docs/design.md
```diff
@@ -0,0 +1,884 @@
...
+| **O8** | **A typed accessor may be wanted later.** §4.1 chose plain dicts. If `spec["content_type"]` becomes a source of typos, the escape hatch with prior art is `pre-commit`'s: a field projection into a frozen dataclass that is structurally incapable of re-validating. Recorded so the answer is not "add pydantic models", which §4.1 rejected for a reason that will still hold |
```
dorado269, line 1144: 小问题记入TODO: P2优先级

agent_sys/examples/demo/agents/compose.jsonnet
```diff
@@ -0,0 +1,14 @@
...
+// declares one even though it never executes anything: a non-leaf's main phase
+// is `unfold`, and `TaskAttempt._main` hands its thread back before a backend is
+// ever selected. `kind: program` is the honest label for "nothing
+// non-deterministic happens here"; there is no `kind: none`, and inventing one
+// would put the axis back where main spec §4.8 took it out of.
+{
+  name: 'compose',
+  kind: 'program',
+  version: '1',
+  description: 'A non-leaf. Its work is its subgraph, so it executes nothing itself.',
```
dorado269: not every task has a agent. a leaf task has agent. non-leaf task has sub-graph.....

agent_sys/examples/demo/closures/describe.jsonnet
```diff
@@ -0,0 +1,36 @@
...
+    // of the model or of the validator.
+    goal: 'Summarise the manifest, including how long the collection took.',
+    version: '1',
+    body: demo.agent_body('describe'),
+    inputs: ['facts'],
+    outputs: ['summary'],
+    permissions: {
+      grants: [
+        demo.grant('facts', 'read'),
+        demo.grant('summary', 'write'),
```
dorado269: an agent automatically have access permission to the task level in/out, so no need declare here?

PR #137, opened 2026-08-25
feat(projection): add InferaSim, a workload-driven serving simulator and projector

jiejingzhangamd:
# Description

Adds `infera/projection/` — InferaSim, a workload-driven simulator and projector for the serving stack. It answers serving-configuration questions without occupying a node to ask each one: time to first token, inter-token latency, throughput, KV and weight memory, feasibility, and cost. One honest experiment on a large model costs a full node for minutes, and these questions are asked in the thousands, so the search happens in simulation and hardware is spent only on the shortlist.

It is deliberately two coupled models over one measured foundation:

- an **analytical projector** that solves a configuration in closed form and answers steady-state questions;
- a **discrete-event simulator** that runs the same cost model on a virtual clock and answers what a closed form structurally cannot — latency *distributions*, queueing under an offered load, and a fleet of replicas sharing a cache and a router.

The governing idea is **measure sparsely, transport analytically**: a small number of cheap sub-scale benchmarks are harvested into anchors, and every other configuration is projected from the nearest applicable anchor rather than measured.

The change is additive. 273 new files under `infera/projection/` and `tests/unit/projection/`, with no existing platform code modified apart from `pyproject.toml`, which gains a `projection` extra and the `inferasim` / `inferasim-tune` entry points.

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- **Analytical inference projection** — forward-only compute for prefill and decode, weight and activation memory per rank, and a KV-cache model covering GQA/MQA, MLA's compressed latent, paged block allocation, sliding-window attention, and KV quantisation.
- **Parallelism** — tensor, pipeline, expert and context parallelism, plus data-parallel attention, the axis MLA models are actually served on: tensor parallelism replicates MLA's latent rather than sharding it, so only splitting by request shrinks the cache a rank holds.
- **Communication model** — TP all-reduce, EP all-to-all and PP point-to-point with selectable algorithms and a batch-dependent compute/comm overlap, priced against measurement rather than assumption, and reported as *exposed* (post-overlap) time.
- **Scheduler and continuous batching** — chunked prefill, token budget, admission delay, and the mixed-step tax that pollutes TPOT, so the projection reflects what goes into each forward pass rather than an idealised one.
- **Discrete-event simulator** — multi-instance fleet behaviour, KV-aware routing scoring the same cost function the deployed router does, a block-level prefix cache, and Mooncake trace replay.
- **Prefill/decode disaggregation** — separate pools with their own parallel shapes and replica counts, charging the KV handoff between them.
- **Benchmark and anchor path** — anchors harvested through the same vLLM/SGLang adapters the platform serves with, so an anchor describes the engine as deployed; an anchor store with regime matching, and a capture ladder for sub-scale measurement.
- **Cost per million tokens** — given a GPU-hour price, throughput reported in the unit a serving budget is quoted in, charging the whole replica so a recipe that buys speed with GPUs is billed for them. Absent a price, absent a figure.
- **Search** — a config sweep that keeps infeasible points annotated rather than dropping them, and a tuning agent that uses the projector as an oracle, warm-started from a deterministic seed sweep.
- **Docs** — `README.md` (task-oriented) and `ARCHITECTURE.md` (how it works, and an explicit statement of what it does not model).

# Checklist:

- [x] New and existing unit tests pass locally with my changes

152 unit tests under `tests/unit/projection/`, all of which run without a GPU.

Made with [Cursor](https://cursor.com)


jiejingzhangamd:
Closing: opened from the wrong account. Reopening as araina-amd, who authored the work.

PR #154, opened 2026-09-03
feat(agent_sys): AgentsView as the o11y panel, one project per run

dorado269:
Wires [AgentsView](https://github.com/kenn-io/agentsview) into `agent_sys` as its observability panel, and groups each run's sessions into one project on it.

**AgentsView's own code is never modified.** Every knob turned is one it already publishes — `--port`, `CLAUDE_PROJECTS_DIR`, `AGENTSVIEW_DATA_DIR`, `disabled_agents`, and its settings API.

## What you get

Deploying `agent_sys` installs and starts the panel on `http://127.0.0.1:18888` (configurable, `--no-agentsview` to skip), leaves it resident across runs, and prints the URL:

```
      o11y  panel at http://127.0.0.1:18888
      o11y  this run is project 'run.20260904T121032_b65238' on the panel
```

## The panel shows only `agent_sys`'s own sessions

Five gates, the strongest of which was found by measuring rather than designing:

| | |
|---|---|
| 1 | `CLAUDE_CONFIG_DIR` in the **child's** environment dict, never in ours. `material.deploy` sets its own per-attempt value, so `<zone>/config/projects` is symlinked into the prefix — credentials stay the zone's, only the output is shared |
| 2 | `CLAUDE_PROJECTS_DIR` points the panel at that one root |
| 3 | `disabled_agents` switches off the other 60 providers, pinned and validated against the installed binary in both directions |
| 4 | A separate `AGENTSVIEW_DATA_DIR`, so a user's own archive is untouched |
| 5 | **`HOME` redirected into the prefix.** AgentsView derives every provider's *default* root from it, so this scopes providers we have never heard of, with no list to go stale |

Verified against the live daemon: `doctor sync` lists **122 agent roots, 122 inside the prefix, 0 outside**.

## One project per run

AgentsView names a project after the session's **deepest** path segment, and every attempt runs in its own nested zone — so one run arrived as several unrelated projects. #156 made those names readable; it cannot join them, because a nested child task fragments off from its parent however prettily both are named, and the only filesystem fix would be putting every attempt under one directory, which is what zone isolation exists to prevent.

So one `explicit` mapping over the run root, posted at run start. **Acceptance was judged on the rendered dropdown**, not an exit code and not the API:

```
[option] run.20260904T121032_b65238 (9)     <- this run
[option] probe (2)
[option] task.0cc49925_...f6daeb1b (1)      <- old behaviour
 … 42 more, every one of them (1)
```

One per-run project holding 9 sessions against 43 holding 1 each — both behaviours in the same control, so it is a comparison rather than an assertion. `demo2` exited 0 with 16 verdicts PASS.

## Measured before it was written, and three of them changed the code

The whole mapping design went through a container recon first:

- **`Origin` is mandatory** on mutating calls — without it the answer is a plain-text `403`, which reads exactly like a missing endpoint.
- **`machine` is read from the daemon's `local_machine`**, never assembled — a wrong one matches nothing, silently, and the recon itself ran in a container whose hostname was not the host's.
- **Names normalise `-` → `_`**, so we normalise before posting and the string we send is the string the panel shows.
- `explicit` is depth-independent; `409` is success; a mapping created before ingest labels sessions automatically, so no `apply`/`preview`/`reclassify`/token.
- Scaling is `O(rows × sessions)` — visible only by repeating the curve at 4× the sessions, where the same row counts cost 3.5–4× more. The path the daemon actually runs is flat: **+32 ms at 5000 rows**, inside the noise.

## Also fixed here, and not in the original scope

**A plain `pytest` was writing 128 MB into the operator's `~/.infera_agent_sys`** — a real 45 MB download and a 133 MB binary — because three test files reached `main(["run", ...])` without patching `ensure_installed`. Nothing said so, and it made the suite need the network. Fixed structurally in two conftests; a scratch prefix now takes 0 bytes.

The readiness probe also ran in the caller's cwd, filing ten identical `Reply with exactly one word: ready` sessions under whichever repository you were standing in. It gets its own directory in the prefix.

## Open, and deliberately not closed

Each is documented with the command that would settle it:

- Why `repo_dot_worktrees` matches nothing across thirteen prefix shapes. We use `explicit`.
- Whether `enabled=false` rows still cost a full-sync scan. Only matters if a GC is ever built; we do not build one, because deleting a row silently un-names that run at the next full sync.
- The zone symlink's behaviour under an enforcing policy — untested because `AGENT_SYS_NO_PERMISSIONS=0` currently fails for every AI package, which predates this work. **Read as "untested", never as "safe".**
- Filed separately as TODO 4g: the backend's `claude` child processes do not exit when their task completes.

**2286 passed, 3 skipped, 4 xfailed.**


.claude/CLAUDE.md
```diff
@@ -1,107 +1,107 @@
-# Task — `e2e_deploy_standardized`: the first stage of the LLM e2e optimisation task package
+# Task — wire **AgentsView** into `agent_sys` as its o11y panel
```
dorado269: do not modify this file when submit to git

agent_sys/cli/environment.py
```diff
@@ -282,6 +283,30 @@ def confinement(availability: Availability | None = None) -> str:
...
 
+def _probe_environment() -> dict[str, str]:
+    """The ambient environment **plus** the o11y prefix's `CLAUDE_CONFIG_DIR`.
+
+    **The probe is a `claude` child too, and it was writing to the user's
+    directory.** Gate 1 of the o11y session scoping covers *agent* children,
+    via `assignment.environment`; this subprocess is not one and so was never
+    covered. It inherited the ambient environment and dropped one JSONL into
+    `~/.claude/projects` on every single run — measured during acceptance by
+    matching the prompt string above to the file. Small, and still a breach of
```
dorado269: comments is too looooong, keep every not file level comment in at most 5 lines, better 3 line

agent_sys/env_mgr/o11y/agentsview.py
```diff
@@ -0,0 +1,675 @@
...
+    if not outs:
+        log.warning(
+            "agentsview: recipe item 'agentsview' not found in %s; skipping the o11y panel.",
+            RECIPE_PATH,
+        )
+        return Status(False, "recipe item not found")
+
+    outcome = outs[-1]
+    if outcome.level == "ok":
+        # **Validated here, at install time, not left for `serve` to discover
```
dorado269: Answer: brief me what this file does

agent_sys/env_mgr/README.md
```diff
@@ -798,6 +798,119 @@ name→path table; it expands nothing.**
...
+`disabled_agents` in a `config.toml` written into our own data directory.
+
+### The prefix, and the variable that names each part of it
+
+The binary and all its state live under one directory `agent_sys` owns, laid out
+like `~/.local`. Nothing is installed anywhere else:
+
+| variable | path | holds |
+|---|---|---|
+| `AGENT_SYS_HOME` | `~/.infera_agent_sys` | the root. Set it to relocate all of the below |
```
dorado269: make this section a single readme file in o11y folder. delete it here(change to oneline brief of under O11y h2 section)

agent_sys/env_mgr/o11y/prefix.py
```diff
@@ -0,0 +1,183 @@
...
+            d.mkdir(parents=True, exist_ok=True)
+
+
+def agent_environment(
+    prefix: Prefix, base: Mapping[str, str], *, bin_on_path: bool = True
+) -> dict[str, str]:
+    """`base`, plus the prefix, plus the one variable that scopes the panel.
+
+    **`CLAUDE_CONFIG_DIR` goes in the returned dict and never into
+    `os.environ`.** That distinction is the whole promise to the user: a Claude
```
dorado269: I think the prefix machinery is a env_mgr level thing, should not lives here o11y folder

work.checkpoint.summary.md
```diff
@@ -1,2690 +1,1149 @@
...
-Reporter reads, cheapest first: the five `*.debug.help.info.md` at the repo
-root; `git log`/`git status` in the worktree; the run roots under
-`/shared_nfs/yihou/agent_sys_debug/ws2/runroot/`; the deliverable dirs; and
-`squeue -u yihou`.
+The previous effort's log is preserved at
+`work.checkpoint.summary.e2e_deploy.20260903-0736.md.bak`.
+
+Reporter reads, cheapest first: `git log --oneline` in the worktree;
+`/home/yihou/ws.agentsview_o11y/recon/PHASE0.md` and `ACCEPTANCE.md`;
+the plan's checkboxes in `docs/superpowers/plans/2026-09-03-agentsview-o11y.md`;
```
dorado269: delete this file

docs/superpowers/specs/2026-09-03-agentsview-o11y-design.md
```diff
@@ -0,0 +1,349 @@
...
+   prefix, **and the process's own `os.environ` is unchanged** — this is the
+   assertion that guards "the user's Claude Code is unaffected".
+5. Port resolution order: flag beats env var beats `18888`.
+
+## 7. End-to-end acceptance
+
+One full `agent-sys run` over `examples/demo2`. Each criterion names a file to
+open and a condition that fails:
+
+| # | Check | How it fails |
```
dorado269: delete, extract the main point into env_mgr's design.md chapter o11y, section agentview

docs/superpowers/plans/2026-09-03-agentsview-o11y.md
```diff
@@ -0,0 +1,1409 @@
...
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Deploying `agent_sys` automatically brings up an AgentsView web panel on port 18888 that shows `agent_sys`'s own agent sessions and nothing else, and that can never fail the deployment.
+
+**Architecture:** AgentsView is installed as an unmodified external binary into a dedicated user prefix `~/.infera_agent_sys` owned by `env_mgr`. Agent child processes get `CLAUDE_CONFIG_DIR` pointed into that prefix, so their transcripts land there instead of `~/.claude`; AgentsView is pointed at that one directory and every other session provider is switched off. A small supervisor module starts the daemon once per deployment and degrades to a single log warning on every failure path.
+
+**Tech Stack:** Python 3 (stdlib `argparse`, `socket`, `subprocess`, `pathlib`, `shutil`), pytest, the existing `env_mgr` recipe/installer machinery, AgentsView (Go binary, external).
+
+**Spec:** `docs/superpowers/specs/2026-09-03-agentsview-o11y-design.md`. Read it first.
```
dorado269: delete this file from git

work.checkpoint.summary.e2e_deploy.20260903-0736.md.bak
```diff
@@ -0,0 +1,2690 @@
...
+never revised, including their wrong estimates — the record over time *is* the
+value of this file.
+
+Effort start (T+0) taken as **2026-09-02 08:28 UTC**, the minute the five
+deliverable dirs under `/shared_nfs/yihou/agent_sys/debugging/` and the ws2
+runroot were created.
+
+Reporter reads, cheapest first: the five `*.debug.help.info.md` at the repo
+root; `git log`/`git status` in the worktree; the run roots under
+`/shared_nfs/yihou/agent_sys_debug/ws2/runroot/`; the deliverable dirs; and
```
dorado269: delete this file

agent_sys/env_mgr/o11y/agentsview.py
dorado269: **`o11y/agentsview.py` — the supervisor for the AgentsView side-car.** ~580 lines, no state of its own, and the one law is that **every function returns a `Status` and raises nothing**: an observability panel that can break the thing it observes is worse than no panel.

Five things:

| | |
|---|---|
| **Port** | `resolve_port` (flag → `AGENTSVIEW_PORT` → 18888, unusable values warn and fall back) and `port_is_free`, a real `bind` rather than a connect. We decide the port because `serve` would otherwise quietly move to the next free one, and a panel on 18889 is a panel nobody knows the address of. |
| **Start / reuse** | `ensure_running`. On a busy port, reuse needs two gates: it answers `/api/v1/agents` with 200 + JSON, **and** a live `daemon.<pid>.json` in *our* `AGENTSVIEW_DATA_DIR` names that port. A user's own AgentsView writes that record into their own data dir, so it structurally cannot appear in ours — that is what stops us handing back a panel listing their whole machine. Otherwise `serve --background --no-browser --replace`, then wait for health. |
| **Config** | `write_config` writes our `config.toml` by rename: `disabled_agents` (the 60 other providers), loopback host, `daemon_idle_timeout = "0s"` so the panel is still there an hour after the run. |
| **Denylist drift** | `OTHER_PROVIDERS` is pinned and hand-maintained; `check_disabled_agents` diffs it against the installed binary in both directions — a name it dropped breaks `serve` loudly, a name it gained leaks silently onto the panel. Via `doctor sync`, never `health`, which autostarts a daemon on a port AgentsView picks. |
| **Install** | `ensure_installed` runs the recipe item through a callable the CLI injects — spec §9 walls the installer machinery off from everything under `env_mgr/`. |

Rationale and measurements live in `env_mgr/docs/design.md` §17.1 and `o11y/README.md`; the file itself now just points at them.

agent_sys/cli/main.py
```diff
@@ -226,6 +254,111 @@ def main(argv: Sequence[str] | None = None) -> int:
...
 
+#: `BinInstaller.install` says "installed <name>" when it ran and "<name>
+#: already present (skip)" when it did not — the only signal `ensure_installed`
+#: passes back about which. `test_the_installers_two_ok_messages_still_discriminate`
+#: drives both real branches, so a rephrasing there fails rather than silently
+#: muting the notice or firing it every run.
+_FRESHLY_INSTALLED_PREFIX = "installed "
+
+
+def _was_freshly_installed(reason: str) -> bool:
```
dorado269: this func should live in env_mgr, utils sub folder or somewhere

agent_sys/cli/main.py
```diff
@@ -226,6 +254,111 @@ def main(argv: Sequence[str] | None = None) -> int:
...
+    return reason.startswith(_FRESHLY_INSTALLED_PREFIX)
+
+
+def _pinned_version() -> str:
```
dorado269: this func should live in env_mgr, utils sub folder or somewhere

agent_sys/cli/main.py
```diff
@@ -226,6 +254,111 @@ def main(argv: Sequence[str] | None = None) -> int:
...
+        _target, items = load_recipe(RECIPE_PATH)
+        for item in items:
+            if item.spec.get("name") == "agentsview":
+                return item.version or "?"
+    except Exception:  # noqa: BLE001 — a notice may not fail what it narrates
+        pass
+    return "?"
+
+
+def _install_item(prefix: Prefix) -> Callable[[], Sequence[Any]]:
```
dorado269, line 264: this func should live in env_mgr, utils sub folder or somewhere

agent_sys/cli/main.py
```diff
@@ -226,6 +254,111 @@ def main(argv: Sequence[str] | None = None) -> int:
...
+
+        target, items = load_recipe(RECIPE_PATH)
+        target.path = str(prefix.root)
+        outs, _status = run(target, items, "install", Filters(item="agentsview"))
+        return outs
+
+    return call
+
+
+def _start_o11y(
```
dorado269, line 302: this func should live in env_mgr/o11y

agent_sys/env_mgr/o11y/agentsview/agentsview.py
```diff
@@ -0,0 +1,582 @@
+# SPDX-License-Identifier: MIT
+# Copyright (c) 2026, Advanced Micro Devices, Inc. All rights reserved.
+"""AgentsView, started as a side-car and never allowed to fail a run.
```
dorado269, line 3: add a sub-folder, move agantsview.py and mapping.py into it

agent_sys/env_mgr/o11y/README.md
```diff
@@ -0,0 +1,145 @@
+# The o11y panel: AgentsView
```
dorado269: move into agentsview subfolder

agent_sys/env_mgr/o11y/README.md
```diff
@@ -0,0 +1,145 @@
...
+
+`agent_sys` ships an observability panel — [AgentsView](https://github.com/kenn-io/agentsview),
+an external Go binary — that reads the session transcripts a run produces and
+serves search, analytics and token-cost views over them at
+**`http://127.0.0.1:18888`**. Deploying `agent_sys` starts it; it stays resident
+across runs, and a later `agent-sys run` only health-checks it and prints the
+URL. It binds loopback only and is not published beyond it.
+
+Design: `../docs/design.md` §17.1.
```
dorado269: 1. design should also write its main body into o11y/agentview/design.md, and the env_mgr/design.md(module level design) should refer to it. not now: put it in env_mgr/design.md and refer back.
2. readme.md should be brief enough, all details lives in design.md

agent_sys/cli/main.py
dorado269: Moved — it is `freshly_installed` / `pinned_version` in `env_mgr/o11y/agentsview/agentsview.py` now, beside the `ensure_installed` whose message they read.

One note on `pinned_version`: it reads the recipe YAML directly instead of calling `recipe.load_recipe`, because that loader is **below the decoupling wall** and nothing under `env_mgr/` may import it. `tests/env_mgr/test_imports.py` derives its "above" set from the filesystem and walks it with `rglob`, so the new subpackage is checked from the moment it exists. One field did not seem worth an exemption.

agent_sys/cli/main.py
dorado269: Moved — it is `freshly_installed` / `pinned_version` in `env_mgr/o11y/agentsview/agentsview.py` now, beside the `ensure_installed` whose message they read.

One note on `pinned_version`: it reads the recipe YAML directly instead of calling `recipe.load_recipe`, because that loader is **below the decoupling wall** and nothing under `env_mgr/` may import it. `tests/env_mgr/test_imports.py` derives its "above" set from the filesystem and walks it with `rglob`, so the new subpackage is checked from the moment it exists. One field did not seem worth an exemption.

agent_sys/cli/main.py
dorado269: **Not moved, and I would rather say why than move it quietly.** `_install_item` exists in `cli/` *because* of the wall it would cross.

`env_mgr` spec §9 forbids anything under `env_mgr/` from importing `recipe`, `runner` or `installers`, and `tests/env_mgr/test_imports.py` enforces it structurally — it derives the "above" set from the filesystem and walks it with `rglob`, with `env_mgr/cli.py` the single exemption. `_install_item`'s whole body is `load_recipe` + `runner.run`, so putting it anywhere under `env_mgr/o11y/` fails that test. A first draft of `ensure_installed` did exactly this and failed exactly that test, which is why the function is injected as a zero-argument callable rather than looked up.

So there are three options and I do not think it is mine to pick:

1. **Leave it in `cli/main.py`** (today). The wall stays intact; the cost is that one o11y-shaped function lives with the CLI.
2. **Move it to `env_mgr/cli.py`**, the one module already exempt. Legal, but that file is env_mgr's own command-line entry point, not a utility home — it would be there for the exemption rather than because it belongs.
3. **Widen the exemption** to a new `env_mgr/o11y/install.py`. Cleanest to read, and it weakens a guard that was put there deliberately — the test's own docstring records a case where a stale hand-written list let a module go unchecked.

Happy to do 2 or 3 if you have a preference; I did not want to weaken the wall on my own judgement.

agent_sys/cli/main.py
dorado269: **Agreed in principle, not done in this push — it is a refactor rather than a file move, and I would rather you saw the shape first.**

`_start_o11y` is bound to `cli/` at two points, and only one of them is incidental:

- it calls `_install_item(prefix)`, which cannot leave `cli/` for the reason in the thread above (it crosses the decoupling wall);
- it emits `EventKind.O11Y_PANEL` on the `Stream`, so moving it as-is would have `env_mgr` importing from `cli` — a library importing its own consumer, which is worse than the thing being fixed.

The honest version is `start_panel(prefix, port_flag, install_item, announce)` in `o11y/agentsview/`, where the caller supplies both the wall-crossing closure and an `announce(message, **fields)` callback, and `cli/main.py` keeps a four-line adapter. That is a good shape. It also re-points about ten tests that currently patch `cli_main.ensure_installed` / `cli_main.ensure_running` — module-level patches that stop reaching the code once it lives elsewhere — so it is a change I would want to make deliberately rather than fold into a review-fix push.

Say the word and I will do it as its own commit. If you would rather keep the blast radius small for now, the other six comments are addressed in `e1490385` and this one can be a follow-up.

PR #155, opened 2026-09-04
[Draft] agent_sys with agent config

dorado269:
# Description

Please include a brief summary of the changes, relevant motivation and context.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

Please list the changes introduced in this PR:

- Change A
- Change B

# Checklist:

- [ ] The functionality is complete
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works


agent_sys/examples/env_checker/steps/check.yaml
```diff
@@ -0,0 +1,283 @@
...
+  # Measured 2026-09-03 (`PROBES.md` D): `uv tool install
+  # "git+https://github.com/oraios/serena"` returns rc 0 and provides `serena`,
+  # `serena-agent`, `serena-hooks`. **`HOME` is deliberately not set here.**
+  # serena's own state also wants redirecting, and probe E did that by pointing
+  # `HOME` at a scratch directory *for the server subprocess* — overriding
+  # `HOME` for the whole agent session would move Claude Code's own state too.
+  # That redirection belongs to the L1 recipe's `env`, not to this block.
+  env:
+    ENVCHK_NONCE: '${nonce}'
+    UV_TOOL_DIR: '${uv_root}/tools'
```
dorado269, line 29: Answer me: So is this a way to declare shell variable for a agent? It nice

agent_sys/env_mgr/addons/envchk-baseline/.claude/servers/envchk_baseline_server.py
```diff
@@ -0,0 +1,173 @@
+#!/usr/bin/env python3
+"""`envchk_baseline` — one MCP tool over stdio, standard library only.
```
dorado269: the whole components folder should lives in the env_mgr folder as a sub-folder, as it a something about env

agent_sys/env_mgr/addons/envchk-baseline/.claude/servers/envchk_baseline_server.py
```diff
@@ -0,0 +1,173 @@
...
+                        "name": TOOL_NAME,
+                        "description": TOOL_DESCRIPTION,
+                        "inputSchema": TOOL_SCHEMA,
+                    }
+                ]
+            },
+        )
+
+    if method == "tools/call":
+        called = request.get("params", {}).get("name")
```
dorado269, line 120: do not implement MCP from scratch
use standard lib
import httpx2
from mcp.server import MCPServer

# Initialize MCPServer
mcp = MCPServer("xxx")

agent_sys/env_mgr/addons/envchk-baseline/.claude/servers/envchk_baseline_server.py
```diff
@@ -0,0 +1,173 @@
...
+
+def _ok(request_id: object, result: dict) -> dict:
+    return {"jsonrpc": "2.0", "id": request_id, "result": result}
+
+
+def _err(request_id: object, code: int, message: str) -> dict:
+    return {"jsonrpc": "2.0", "id": request_id, "error": {"code": code, "message": message}}
+
+
+def main() -> int:
```
dorado269, line 139: Answer me: So every time the mcp call is invoked by claude code, a new python process is start to response here?

agent_sys/examples/env_checker/assets/env_probe.agent/.claude/.mcp.json
```diff
@@ -0,0 +1,17 @@
+{
+  "mcpServers": {
+    "serena": {
+      "type": "stdio",
+      "command": "${UV_TOOL_BIN_DIR}/serena",
```
dorado269, line 13: I think there is an official claude code plugin can install serena, so why this file exists?

agent_sys/spec_loader/assets.py
```diff
@@ -270,6 +326,57 @@ def fill_body(
...
+    *,
+    kind: str,
+    name: str,
+    origin: str,
+    line: int | None,
+) -> list[Problem]:
+    """Fill an agent's `assets` from the index, or warn that it was bound by hand.
+
+    **Only `kind == "agent"`.** Every other kind returns `[]` untouched, because
+    `assets` is not in their schemas and writing it would put a key there that
```
dorado269: really? not sure, but it seems all materials of every type[handoff、task、validator] live in the "assert" root with sub-folders. and the sub-folder resolve way is already implemented. so why again finding assets implemention? Am I missing something? so how did validators like "check_env_report_shape.validator" find their folder?

agent_sys/examples/env_checker/steps/check.yaml
```diff
@@ -0,0 +1,283 @@
...
+  #   **nothing in the package declared serena as an MCP server**. The
+  #   invocation existed only in a comment in `env_mgr/recipes/serena.yaml`, and
+  #   a comment is not a declaration.
+  #
+  # **`recipes: [serena]` and `components: [serena]` are both required and that
+  # is not redundant.** Install and declaration are two halves; neither implies
+  # the other, and run 1 is what proved it. Capability 7 is therefore
+  # **installed at L1 and declared at L2** — a genuine split, written down in
+  # `ACCEPTANCE.md` row 7 rather than smoothed over.
+  components: [envchk-baseline, serena]
```
dorado269: components should also declared in the recipes.

agent_sys/examples/env_checker/steps/check.yaml
```diff
@@ -0,0 +1,283 @@
...
+  # **L1 — an industry component, installed by an `env_mgr` recipe.** serena is
+  # a real third-party language server bundle; `recipes:` is the existing recipe
+  # machinery (`env_mgr/recipe.py` + `installers/`), which until this round was
+  # reachable only from the `env-mgr` CLI and never ran during a task at all.
+  # This is the first package to name one.
+  #
+  # It is one name and not two on purpose. The point of L1 here is the *route*,
+  # and a second recipe would add a second way for the run to fail on somebody
+  # else's network without measuring anything the first does not.
+  recipes: [serena]
```
dorado269: recipes can also be defined in task_package.
and rename 'recipe[s]' to "env_recipe[s]"
1. env_recipe.${agent_name}.yaml or ${agent_name}.env_recipe.yaml is automatically detected in the whole test package.
2. the task_package can have it own main.env_recipe.yaml or env_recipe.main.yaml . as the workspace level[kind] (If i am not wrong, there is a overlay/fallback system of different kind recipe like workspace/repo/project..., tell me details about that and we make sure how next step do.)
3. embedded recipes define is not allowed. but you can specifically refer to an exist recipe. 
4. I don't think list/arr is needed here.

any unclear terms, please ask me.

agent_sys/examples/env_checker/steps/check.yaml
```diff
@@ -0,0 +1,283 @@
...
+  #
+  # Measured 2026-09-03 (`PROBES.md` D): `uv tool install
+  # "git+https://github.com/oraios/serena"` returns rc 0 and provides `serena`,
+  # `serena-agent`, `serena-hooks`. **`HOME` is deliberately not set here.**
+  # serena's own state also wants redirecting, and probe E did that by pointing
+  # `HOME` at a scratch directory *for the server subprocess* — overriding
+  # `HOME` for the whole agent session would move Claude Code's own state too.
+  # That redirection belongs to the L1 recipe's `env`, not to this block.
+  env:
+    ENVCHK_NONCE: '${nonce}'
```
dorado269, line 28: what is nonce

agent_sys/env_mgr/recipes/serena.yaml
```diff
@@ -0,0 +1,160 @@
+# SPDX-License-Identifier: MIT
```
dorado269, line 1: add a default.yaml recipe as the very default recipe if nothing defined for an agent in task package(no main recipe, no agent recipe).
1. use most bottom level kind
2. please add my current user claude plugins into it (see what I installed and use claude intsall ...)
3. add cli
4. install something else, make it good enough for claude code to works for a develop and experiment job working in repos like infera/sglang.
5. print warning when using it for an agent

agent_sys/examples/env_checker/steps/check.yaml
dorado269: components is a tag identify that this item is from inside the agent_sys.
the there L1/L2/L3 classify way is a virtual concept not a user aware thing

agent_sys/env_mgr/agent_assets.py
```diff
@@ -0,0 +1,1655 @@
...
+| L1 | industry components: serena, a marketplace plugin, an apt/pip tool | ``recipes: [...]`` -> an `env_mgr` recipe YAML |
+| L2 | components this repository ships | ``components: [...]`` -> `agent_sys/components/<name>/` |
+| L3 | components one task package carries for one agent | **undeclared** — `<agent assets>/.claude/` |
+
+**L2 and L3 have the same on-disk shape**, so one installer serves both and
+promoting a component is moving a directory. **L3 is undeclared** because the
+declaration would be a second statement of what the directory already says, and
+the two would drift; `components/README.md` states the same contract from the
+author's side.
+
```
dorado269: Answer me:what's the relationship of this file and material.py

agent_sys/examples/env_checker/steps/check.yaml
dorado269: actually I think the L2 should not exist

and L1/L3 should also not exist in the core code.
just reuse the recipe level system to implement it. make L3 as a new top layer or a repo layer

dorado269: more concept make more complexity to maintain

dorado269: L1 L3 is just a user friendly declare way, not a new system for the core code

agent_sys/env_mgr/recipes/serena.yaml
dorado269: sorry, add gh cli

agent_sys/examples/env_checker/steps/check.yaml
dorado269: Yes. `env:` on the step is the agent process's environment — every key becomes a shell variable the agent and anything it spawns can read. Values interpolate `--var` substitutions, so `ENVCHK_NONCE: '${nonce}'` binds the run's nonce into the session.

It is per-step and per-agent, not global: a variable declared here does not reach another step. That is why serena's state redirection is not here but on its own `.mcp.json` entry — a session-wide variable would apply to everything the agent runs, which is wider than the one process that needs it.

agent_sys/env_mgr/addons/envchk-baseline/.claude/servers/envchk_baseline_server.py
dorado269: Done. `components/` is now `agent_sys/env_mgr/addons/`, inside `env_mgr` as you asked, and the directory keeps its contents. `env_mgr/addons/README.md` states the contract from the author's side.

There is no declaration key that reaches it. A recipe that needs an add-on locates it by importing `env_mgr`, so the path is never exported to the zone.

agent_sys/env_mgr/addons/envchk-baseline/.claude/servers/envchk_baseline_server.py
dorado269: Not done, and I would rather say so than quietly close this. You are right about the shape — hand-rolled JSON-RPC is the wrong thing to ship.

What blocks it is the interpreter, not the code. These servers are launched as bare `python3 <script>` from inside a zone, so the interpreter is whichever `python3` the zone's `PATH` selects — a system one, which has no `mcp` package. Verified: `/usr/bin/python3 -c 'import mcp'` raises `ModuleNotFoundError`. The failure mode is the bad one — the server exits at import and Claude Code reports **zero tools rather than an error**, with nothing in the run naming the cause.

Closing it means either having the recipe layer install `mcp` into the zone and pinning the server's interpreter to it, or launching servers through a resolved interpreter rather than a bare name. Both are wider than rewriting the two files, so it is recorded as `docs/TODO.md` 4m rather than guessed at. Leaving this thread open for you to rule on which route.

Edit: the row is **4m**, not 4k — 4k was already taken. Its second claim is also marked as inferred there rather than measured: the missing `mcp` package is measured, but that an `ImportError` at start-up presents as zero tools rather than an error has not been driven for this case.

agent_sys/env_mgr/addons/envchk-baseline/.claude/servers/envchk_baseline_server.py
dorado269: No — one process per session, not one per call. `main()` is `for line in sys.stdin:`, so the process starts once when the session opens and reads a stream of line-delimited JSON requests until stdin closes. Every tool call in the session is handled by that same process.

That is the `"type": "stdio"` contract, and it is why the entry writes `command` and `args` rather than a URL: Claude Code owns the process lifetime. A port-based server, which does outlive the session, goes through the `run_server` installer instead.

agent_sys/examples/env_checker/assets/env_probe.agent/.claude/.mcp.json
dorado269: Because the official plugin's entry cannot express two things this run needs.

As measured when this was written, that entry carries no `env` block and no `--project`, and `--project` was found to have no environment-variable equivalent in serena's CLI — so it can only be passed as an argument. If the plugin has gained either since, this file should go and I would rather delete it than keep it; say so and I will re-measure against the current plugin.

The part of your objection that did land: we were overriding `HOME` to relocate serena's state, which is far wider than the one process. That is now `SERENA_HOME`, which covers everything serena writes under `$HOME` and touches nothing else.

agent_sys/spec_loader/assets.py
dorado269: You are not missing anything — you are describing the defect.

`assets.py` resolves **one file per role** (`body.readme`, `body.entry`) by filename convention, scoped by an optional `<name>[.<type>]/` folder. That is how `check_env_report_shape.validator` finds its folder. Every *other* file an object needs is carried by nothing: they arrive because `layout.stage_package` copies the whole package into the zone, and each body then reaches them by a hand-built path. So `body.entry` is a pointer, not a manifest, and the object's resource set is never named anywhere.

Recorded as `docs/TODO.md` 4g — a small-scope refactor, not a rewrite. Resolving this thread since the gap is now written down where the next person will find it.

agent_sys/examples/env_checker/steps/check.yaml
dorado269: Done — the level model is gone. There is no L1/L2/L3 anywhere, no `layer:` key in a recipe, and no `layer` field on an item; a stale `layer:` is now rejected rather than ignored.

What replaced it is the recipe layer system you pointed at. Three layers, and the layer is **where the file is**, never a field: `env_mgr/default.env_recipe.yaml` (always applies, declared by nobody), `<package>/assets/main.env_recipe.yaml` (auto-detected), and the agent's own `recipes:`. They concatenate, most general to most specific.

Dropping the field was right for a reason beyond the extra concept: who declared an install is already carried by the file's path, so the field was a second writer of one fact.

agent_sys/examples/env_checker/steps/check.yaml
dorado269: Three of the four are done; the fourth is not, and I have left this open for that.

**1. Auto-detection — done.** `env_recipe.<agent>.yaml`, `<agent>.env_recipe.yaml` and any other `.`-joined permutation are discovered, including inside the agent's own `<name>.agent/` folder where the name may be dropped. Scoped to `assets/`: a recipe at the package root is a load failure. Two spellings for one agent is `SpecInconsistent`.

**2. The task package's own recipe — done.** `<package>/assets/main.env_recipe.yaml`, that exact filename only, auto-detected and never declared. On the overlay question you asked about: the layers **concatenate**, they do not override. A later layer adds items; it never replaces an earlier one. That costs nothing because every installer gates on `check` before `install`, so an item two layers both declare is done once and reported twice.

**3. Embedded definitions — done.** `recipes:` accepts references only; an inline recipe object fails schema validation. Every reference names its root — `agent_sys:<name>` or `package:<relpath>` — and a bare name is refused with both spellings named.

**4. The list — not done.** `recipes:` is still an array. One agent can legitimately want more than one recipe, and collapsing to a single value would remove that with no replacement. But that is my reasoning, not your ruling, so I have not made the call: tell me whether you want a scalar and I will change it.

agent_sys/examples/env_checker/steps/check.yaml
dorado269: A per-run random value, supplied as `--var nonce=...`, that makes a capability report unforgeable.

Each capability's token is `sha256(salt:label:nonce)[:12]`. The salt lives in exactly one authored file per capability, so an agent can only produce the right token by actually reaching that capability. The nonce is what stops the answers being reusable: without it the tokens would be constants, and the first published handoff would contain the answers for every run after it.

It deliberately has no default, so a run that forgets it fails at load time naming the file and line rather than silently producing a constant.

agent_sys/env_mgr/recipes/serena.yaml
dorado269: Partly done. The layer exists; its contents do not yet match what you asked for, so I am leaving this open.

**Done:** `env_mgr/default.env_recipe.yaml` is the default layer. It is at the bottom kind (`workspace`), it applies to every agent, nothing declares it, and its absence is simply absence. It sits in `env_mgr/` rather than `env_mgr/recipes/` so that the path says it is the default and no field has to.

**Not done, and each needs a decision from you:**
- **`gh` and a wider CLI set.** I did not add an install command I cannot exercise here — `gh` already resolves on this class of host, so the install branch would never run and could not be claimed as tested, and a rootless `gh` install is a real choice (distro package vs. release tarball) rather than a one-liner. Tell me which and I will add it.
- **Your marketplace plugins.** Installing them needs `claude plugin marketplace add anthropics/claude-plugins-official` first, and that clone was observed going over SSH. A zone with no forwarded key fails it, and whether the CLI falls back to HTTPS is unmeasured.
- **The warning.** Not implemented, and it interacts with a design point worth naming: you described the default as a *fallback* for an agent with no other recipe, but it is currently a **layer that always applies** and concatenates under the others. Those are different mechanisms and only one should exist. Which did you intend?

agent_sys/env_mgr/agent_assets.py
dorado269: `material.py` places files; `agent_assets.py` places everything that is not a file.

`material.py` deploys what the agent spec declares as three lists of *file paths* — `rules`, `hooks`, `skills` — in Claude Code's canonical form, and it parses nothing: a file is placed, not read.

That shape stops at the first real component. A skill is a directory, a plugin marketplace is a directory of directories, an MCP server is a process to register rather than a file to place. `agent_assets.py` is the route for those, and it also owns the three documents whose *contents* are an interface (`settings.json`, `.mcp.json`, `marketplace.json`) — it has to, because installs merge into `settings.json` rather than clobbering it, so the file must exist first and only this module knows when that moment is.

So `material.py` keeps the four declared keys and calls the one module that owns the rest. Both module docstrings now say this.

PR #156, opened 2026-09-04
feat(agent_sys): label every runtime directory with the name it alrea…

dorado269:
…dy knew

A run tree was browsable only if you already knew the uuids: `handoffs/<uuid>/`, `zones/task.<uuid>.<attempt>.<hash>/`. The closure name and the handoff kind were both in hand at the moment the directory was created, and both thrown away. They are now the second field:

    handoffs/handoff.solutions_a.57882e8c-.../v0/content/
    zones/task.main.2c0260e5-.../task.solve_a.310209e3-....0.ce4e41e0/
    zones/task.main.2c0260e5-.../validation.solve_a.310209e3-....output_validation.01a143f0/

Two properties keep the label from becoming a contract:

- **Nothing resolves through it.** `find_zone_dir` matches the uuid as a whole field, and `handoff_dir` finds the directory whose name *is* the uuid or ends with `.<uuid>`. So a store or zone tree written before labels existed resolves unchanged, and a renamed closure does not move an artefact.
- **`.` does not survive a label.** The uuid is the last field before the attempt/version, and `parts[-2]` would read the wrong one otherwise.

The staged copy in a zone takes the store directory's own basename rather than recomputing the label, which keeps `task_graph`'s kind vocabulary out of `stage`, `stage_handoffs` and `prepare_validation`.

`handoff.store.version_dir` and `env_mgr.fs.layout.handoff_version_dir` stay the two sanctioned writers of the shape; `tests/interfaces/test_handoff_layout.py` pins both, including that either alone would regress the compatibility rule. The five body-side `examples/*/assets/lib/store.py` readers get the same scan-then-fallback rule under the same agreement test.

Verified: 2266 passed / 3 skipped / 4 xfailed; `examples/demo2` run end to end and the tree above read off it; that run's store renamed back to bare uuids and re-read through all three readers.

# Description

Please include a brief summary of the changes, relevant motivation and context.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

Please list the changes introduced in this PR:

- Change A
- Change B

# Checklist:

- [ ] The functionality is complete
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/agent_sys/examples/demo2/assets/lib/store.py
+++ b/agent_sys/examples/demo2/assets/lib/store.py
@@ -58,6 +58,30 @@ def store_root() -> Path:
     return Path(os.environ["AGENT_SYS_DEMO_STORE"])
 
 
+def handoff_dir(hid: str) -> Path:
+    """This handoff's directory in the store. `handoff.store.handoff_dir`, duplicated.
+
+    The store names a directory ``handoff.<kind>.<uuid>`` so that a person
+    reading a run tree can tell what an artefact is. **Nothing resolves through
+    the label**: a directory is this handoff's when its name *is* the uuid — the
+    shape written before labels existed — or ends with ``.<uuid>``. The same
+    admissible duplication as `MANIFEST` and `CONTENT` above, and covered by the
+    same agreement test.
+    """
+    root = store_root()
+    suffix = f".{hid}"
+    if root.is_dir():
+        for entry in sorted(root.iterdir()):
+            if entry.name == hid or entry.name.endswith(suffix):
+                return entry
+    return root / hid
+
+
+def hid_of(dirname: str) -> str:
+    """The handoff id in a store directory name — the last field, labelled or not."""
+    return dirname.rsplit(".", 1)[-1]
+
+
 def versions(hid: str) -> list[int]:
     """**Published versions only, and the gaps are the point.**
 
@@ -75,7 +99,7 @@ def versions(hid: str) -> list[int]:
     next. Holes are skipped and never compacted, because renumbering would move
     an artefact a digest already names.
     """
-    base = store_root() / hid
+    base = handoff_dir(hid)
     if not base.is_dir():
         return []
     found = [
@@ -92,7 +116,7 @@ def content_dir(hid: str, version: int | None = None) -> Path | None:
     if not numbers:
         return None
     chosen = numbers[-1] if version is None else version
-    path = store_root() / hid / f"v{chosen}" / CONTENT
+    path = handoff_dir(hid) / f"v{chosen}" / CONTENT
     return path if path.is_dir() else None
 
 
@@ -101,7 +125,7 @@ def kind_of(hid: str, version: int | None = None) -> str:
     if not numbers:
         return ""
     chosen = numbers[-1] if version is None else version
-    manifest = store_root() / hid / f"v{chosen}" / MANIFEST
+    manifest = handoff_dir(hid) / f"v{chosen}" / MANIFEST
     if not manifest.is_file():
         return ""
     # **YAML, not JSON**, and this read `json.loads` until the filename was
@@ -173,9 +197,10 @@ def latest_of_kind(kind: str) -> Path | None:
         return None
     best: tuple[float, Path] | None = None
     for entry in sorted(root.iterdir()):
-        if not entry.is_dir() or kind_of(entry.name) != kind:
+        hid = hid_of(entry.name)
+        if not entry.is_dir() or kind_of(hid) != kind:
             continue
-        found = content_dir(entry.name)
+        found = content_dir(hid)
         if found is None:
             continue
         stamp = found.stat().st_mtime
--- a/agent_sys/examples/llm_e2e_performance_optimization/analyze-demo/assets/lib/store.py
+++ b/agent_sys/examples/llm_e2e_performance_optimization/analyze-demo/assets/lib/store.py
@@ -53,6 +53,30 @@ def store_root() -> Path:
     return Path(os.environ["AGENT_SYS_DEMO_STORE"])
 
 
+def handoff_dir(hid: str) -> Path:
+    """This handoff's directory in the store. `handoff.store.handoff_dir`, duplicated.
+
+    The store names a directory ``handoff.<kind>.<uuid>`` so that a person
+    reading a run tree can tell what an artefact is. **Nothing resolves through
+    the label**: a directory is this handoff's when its name *is* the uuid — the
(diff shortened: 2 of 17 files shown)


agent_sys/examples/demo2/assets/lib/store.py
```diff
@@ -101,7 +125,7 @@ def kind_of(hid: str, version: int | None = None) -> str:
     if not numbers:
         return ""
     chosen = numbers[-1] if version is None else version
-    manifest = store_root() / hid / f"v{chosen}" / MANIFEST
```
dorado269, line 104: Answer: why this demo test package should know something about agent_sys?

agent_sys/examples/llm_e2e_performance_optimization/analyze-demo/assets/lib/store.py
```diff
@@ -70,7 +94,7 @@ def versions(hid: str) -> list[int]:
     next. Holes are skipped and never compacted, because renumbering would move
     an artefact a digest already names.
     """
-    base = store_root() / hid
```
dorado269, line 73: Answer please

agent_sys/examples/llm_e2e_performance_optimization/integration-demo/assets/lib/store.py
```diff
@@ -53,6 +53,30 @@ def store_root() -> Path:
     return Path(os.environ["AGENT_SYS_DEMO_STORE"])
 
 
+def handoff_dir(hid: str) -> Path:
```
dorado269, line 56: ...

agent_sys/examples/demo2/assets/lib/store.py
dorado269: It doesn't know anything about `agent_sys` the *package* — it never imports one. It knows exactly one fact: **`handoff`'s on-disk store layout**. That is stated at the top of this very file, and it predates this PR — the file has read `<root>/<hid>/v<N>/` since `3f8a15c`, the initial commit of the system.

**Why it has to.** A task package is data. `examples/demo2/README.md`: *"there is no `__init__.py` here and there never will be … it may use nothing a task package outside this repository could not use: no privileged import, no private loader path."* So `lib/store.py` cannot call `handoff.version_dir`, even though it needs the same answer. A validator body runs as a subprocess and is handed `inputs.json` (ids) and `materials.json` (paths to the staged copies of what it is validating). For a handoff it was **not** handed — which is what *"grounded in its input"* requires — it has no route at all, so it reads the store. `docs/interfaces.md` §8.1 admits that duplication on one condition: an agreement test that *fires*.

**That test exists and it fired on this PR.** `tests/cli/test_isolation_shown.py::test_the_store_layout_this_package_reads_is_handoffs` builds its fixture with `handoff.version_dir` and `handoff`'s own constants — deliberately never with this module's — then asserts this module finds it. Under the new layout the fixture is `handoff.<uuid>/v7`, and the test passes. That is the evidence the two readers still agree; it is not an argument.

**It is not dead code**, measured rather than assumed: `content_dir(hid)` is the `or` fallback in ~30 validators across the five packages, and `latest_of_kind` is a live route in `demo2/check_problems.validator` and `demo/check_grounded.validator`.

**What this PR does add, stated plainly.** Before, the file duplicated three constants and the path shape. It now also duplicates the *resolution rule* — "the directory whose name is the uuid, or ends with `.<uuid>`". That is a real widening of the duplicated surface, and it is the fair reading of your question.

The change that would remove it rather than widen it is to delete the store-scanning route entirely and leave `materials.json` and `AGENT_SYS_{INPUT,OUTPUT}_<KIND>` as the only ones. This file's own docstring already calls the scan *"crude"*, *"wrong in a real graph"*, and notes it **cannot work under confinement at all** (measured `EACCES` on the store root from a confined body). That is a behaviour change to ~30 validators, not a naming change, so I did not fold it into this PR. Say the word and I'll open it as a follow-up.

All five copies are byte-identical in the added block (same md5 over `handoff_dir`/`hid_of`), per the "keep it verbatim, fix the original and copy it across" rule at the top of the file.

agent_sys/examples/llm_e2e_performance_optimization/analyze-demo/assets/lib/store.py
dorado269: Same question as on `demo2/assets/lib/store.py` — full answer there: https://github.com/AMD-AGI/Infera/pull/156#discussion_r3932714134

Short form for this line: `base = store_root() / hid` was already this file composing `handoff`'s layout, since the initial commit. It composes it because a task package may not import a component (`README.md`: no `__init__.py`, no privileged import), and a validator handed only `inputs.json` has no other route to a handoff it was *not* handed. `docs/interfaces.md` §8.1 permits that on the condition of an agreement test, and `tests/cli/test_isolation_shown.py::test_the_store_layout_this_package_reads_is_handoffs` is it — it builds its fixture with `handoff.version_dir` and passes under the new layout.

What this PR changes on this line is only *which* directory `base` is: it is now resolved (name is the uuid, or ends with `.<uuid>`) instead of composed, so a store written before labels existed still reads. Verified against a real `demo2` run whose store I renamed back to bare uuids.

agent_sys/examples/llm_e2e_performance_optimization/integration-demo/assets/lib/store.py
dorado269: This is the function your other two comments are about — full answer here: https://github.com/AMD-AGI/Infera/pull/156#discussion_r3932714134

In one line: it exists because a task package is data and may not import `handoff`, so it cannot call `handoff.store.handoff_dir`; `docs/interfaces.md` §8.1 allows the duplicate reader only with an agreement test, and `tests/cli/test_isolation_shown.py::test_the_store_layout_this_package_reads_is_handoffs` builds its fixture from `handoff`'s own functions and passes.

I do concede the surface got wider: the file used to duplicate three constants plus a path shape, and now duplicates a resolution rule as well. The way to shrink it instead is to delete the store-scanning route (`versions`/`content_dir`/`kind_of`/`latest_of_kind`) and leave `materials.json` and `AGENT_SYS_{INPUT,OUTPUT}_<KIND>` as the only ones — the docstrings here already say the scan is crude and does not work under confinement. That touches ~30 validators, so it is a follow-up rather than part of a naming change. Happy to open it.

PR #163, opened 2026-09-10
GLM-5.3 (glm_moe_dsa): mixed, 1P1D, and a Kubernetes recipe — plus overlay support for SGLang bases

dorado269:
## What

Adds GLM-5.3 (the `glm_moe_dsa` "big" family, including `-MXFP4`) as a supported
model, in three shapes:

- **single-node mixed** — `examples/sglang_mix_glm5.3/`
- **1P1D under docker** — `examples/sglang_1p1d_glm5.3/`, driving the GLM-5.2 kit
  rather than forking it
- **1P1D under Kubernetes** — `examples/recipes/glm5.3/disaggregated/deploy.yaml`
  plus `manual/examples/k8s_glm5.3.md`

and, to make the last one possible, teaches the **overlay payload to patch an
SGLang base**.

## Why the overlay change is the load-bearing part

`deploy/overlay/Dockerfile.payload` shipped `patches/vllm/` only, and
`deploy/overlay/infera-exec` gated its runtime patch loop on `import vllm`. On an
SGLang base that gate is closed, so **no engine patch was applied at all**, while
the baked `Dockerfile.sglang` applies six. On gfx950 the two that matter are a
DP-attention host-sync deadlock and an aiter paged-MQA row-sizing bug — the first
hangs, the second returns wrong output behind a 200.

This PR ships the four sglang patch trees plus `apply_sglang_dsa_patches.sh` and
gives `infera-exec` an `elif import sglang` branch. Three of the groups are plain
`.py` anchor scripts, which is already the runtime-patch contract; `sglang_dsa` is
one script plus three `--fuzz=0` diffs, so the applier that knows how to drive
them — and how to verify they reached the **bytecode**, not just the source — is
invoked for that group. It runs under a marker and a flock because an engine leg
and the router both pass through `infera-exec` inside one container.

Two omissions there are deliberate and commented in place: `sglang_carried/` is
not shipped (it perturbs MoE routing, is not GPU-validated, and the baked image
does not apply it either), and `sglang_dsa/` is deliberately not run through the
`*.py` glob (it contains a v0.5.16-only port that would fail noisily on v0.5.18).

## Validation

MI355X, 8× gfx950 per node, GLM-5.3-MXFP4, TP4 per leg, symmetric DP-attention,
mooncake RDMA over RoCE. Three configurations, each moving exactly one axis from
the last:

| | engine image | orchestration | correctness | output tok/s |
|---|---|---|---|---|
| baseline | baked | docker | PASS | — |
| overlay | **stock** `lmsysorg/sglang:v0.5.18-rocm720-mi35x` | docker | PASS | 343.5 |
| k8s | stock | **k3s** | PASS | 348.3 |

"PASS" is the same probe in all three: a short answer, arithmetic, a 41-token
needle, two **repeats**, a 7845-token needle recovered verbatim, and 32 requests
at concurrency 8 with 0/32 degenerate. Token counts and `finish_reason` are
identical across all three rows.

The repeats and the load round are not padding. The failure this configuration can
hit — asymmetric DP-attention under PD — returns a clean **first** completion and
garbage on every later one, with zero faults, zero HIP errors, and nothing in any
log. A single-shot check passes a broken deployment.

## Also in here

- `examples/sglang_1p1d_glm5.2/` gains overlay mode (`OVERLAY_PAYLOAD`,
  `HOST_LIBIONIC`, `INFERA_EXEC`), all defaulting to empty so a baked image
  behaves exactly as before, plus per-leg knobs a single-node pair needs.
- `bench.sh` gains `docker exec -w /`: on a stock base the image WORKDIR puts
  `sglang` on `sys.path[0]` as a namespace package and `bench_serving` dies on
  `No module named 'sglang.benchmark.serving'`.
- `deploy/docker/patches/sglang_carried/` — a GLM MoE gate-bias fp32 fix, carried
  **unwired**: bf16 collapses 238 distinct correction-bias values to 8, which
  reorders top-k routing. Not applied anywhere, because it is a routing
  perturbation rather than a crash and has not been GPU-validated.
- Two GLM-5.3 e2e rows in `tests/e2e/pd_mixed/sglang/matrix.py`, **`enable=False`**
  as the CI contract requires, recipes intact.
- `manual/sphinx/_toc.yml.in` — registers the new page so it appears in the built
  navigation rather than being reachable only by URL.

## CI

`lint` is red, and **none of it is from this branch**. `pre-commit` runs
`--all-files`, and the failures are 83 files under `infera/projection/**` and
`tests/unit/projection/**` that this PR does not touch; the same job is failing on
`main` at `6492209b`. The three Python files this PR does change are clean —
`ruff format --check` and `ruff check` both pass on them.

`e2e-*` are skipped by the repo's own gate. The GLM-5.3 e2e rows added here are
`enable=False` and do not run.

## Scope

**GLM-5.3-Flash is not in this PR.** It is `glm5_next` — a different architecture
that no published SGLang release can load — and it lands separately.

### DIFF
--- a/deploy/overlay/Dockerfile.payload
+++ b/deploy/overlay/Dockerfile.payload
@@ -122,8 +122,54 @@ COPY --from=native312 /usr/local/bin/infera-router /payload/bin/infera-router
 # Engine patches applied at container start (see infera-exec). These fix vendor
 # code we do not otherwise ship — the overlay's whole point is not forking the
 # vendor image, and a runtime patch keeps that true where a rebuild would not.
-# Each script is self-locating, idempotent, and no-ops once upstream carries the
-# fix, so they are safe to leave in place across a base bump.
+# The vllm/ scripts are self-locating and atomic: they verify every anchor before
+# the single write, so a base whose layout moved exits 1 leaving the file
+# untouched. That is what makes THEM safe across a base bump. It is not true of
+# sglang_dsa — see the TODO below.
 COPY deploy/docker/patches/vllm/ /payload/patches/vllm/
+# The SGLang groups, without which an SGLang base overlay serves an entirely
+# UNPATCHED engine: on gfx950 that is a DP-attention host-sync deadlock (a hang)
+# plus an aiter paged-MQA row-sizing bug that returns wrong output behind a 200,
+# so nothing in any log says why. disagg/responses/rocm are plain `.py` scripts.
+COPY deploy/docker/patches/sglang_dsa/       /payload/patches/sglang_dsa/
+COPY deploy/docker/patches/sglang_disagg/    /payload/patches/sglang_disagg/
+COPY deploy/docker/patches/sglang_responses/ /payload/patches/sglang_responses/
+COPY deploy/docker/patches/sglang_rocm/      /payload/patches/sglang_rocm/
+# sglang_dsa is diffs pinned at `--fuzz=0`, not runnable scripts, so its applier
+# ships too and drives that group instead of infera-exec's `<dir>/*.py` glob —
+# the glob would also run a v0.5.16-only script and fail on this base. For the
+# same no-opt-out reason `sglang_carried/` is deliberately NOT copied.
+COPY deploy/docker/scripts/apply_sglang_dsa_patches.sh /payload/bin/apply_sglang_dsa_patches.sh
+#
+# TODO — this payload is VERSION-BOUND TO THE ENGINE, and nothing enforces it.
+#
+# infera-exec resolves four axes at container start. Three declare themselves and
+# fail loudly on a mismatch: python minor (`py310/`), ROCm major
+# (`native/rocm7-py310/` + CAPABILITIES + INFERA_REQUIRE_NATIVE), and engine
+# family (`import vllm` / `import sglang`). The fourth — **engine version** — is
+# not checked at all, and sglang_dsa is `--fuzz=0` diffs cut against exactly one
+# release (today v0.5.18).
+#
+# Measured on `lmsysorg/sglang-rocm:rocm720-mi35x-k3-20260727` (sglang 0.5.15,
+# the Kimi-K3 base): disagg/responses/rocm all apply, then
+# draft_cuda_graph_dp_vote lands 6 of its 7 files and rejects every hunk of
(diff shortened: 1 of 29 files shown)


deploy/overlay/Dockerfile.payload
```diff
@@ -125,5 +125,18 @@ COPY --from=native312 /usr/local/bin/infera-router /payload/bin/infera-router
...
 # fix, so they are safe to leave in place across a base bump.
 COPY deploy/docker/patches/vllm/ /payload/patches/vllm/
+# The SGLang groups, without which an SGLang base overlay serves an entirely
+# UNPATCHED engine: on gfx950 that is a DP-attention host-sync deadlock (a hang)
+# plus an aiter paged-MQA row-sizing bug that returns wrong output behind a 200,
+# so nothing in any log says why. disagg/responses/rocm are plain `.py` scripts.
+COPY deploy/docker/patches/sglang_dsa/       /payload/patches/sglang_dsa/
+COPY deploy/docker/patches/sglang_disagg/    /payload/patches/sglang_disagg/
+COPY deploy/docker/patches/sglang_responses/ /payload/patches/sglang_responses/
+COPY deploy/docker/patches/sglang_rocm/      /payload/patches/sglang_rocm/
```
dorado269, line 137: 1. ensure this does not break kimi k3
2. e2e test support overlay mode

examples/recipes/glm5.3/disaggregated/deploy.yaml
```diff
@@ -0,0 +1,364 @@
...
+#    cannot say WHICH ones a pod gets, and it renumbers the allocation from 0 so
+#    visible-device pinning on top of it silently masks GPUs. Pin with
+#    HIP_VISIBLE_DEVICES + nodeSelector; re-add the requests if you add the plugin.
+#
+# 2. The overlay payload must be built from THIS repo. The published
+#    `inferaimage/infera-overlay` tag carries `patches/vllm/` only and its
+#    infera-exec gates the patch loop on `import vllm`, so on an SGLang base it
+#    applies nothing and the engine runs unpatched. See the README for the build.
+#
+# 3. NO MTP, and TP4 rather than TP8. Both come from the validated run:
```
dorado269: ??

jiejingzhangamd (approved):

